PR #13, opened 2024-01-19
Improve Eq Terminology

### LINKED ISSUE #10: Use consistent terminology for Eq's properties
The terminology we use in the BIP should try to both be clear and consistent with existing literature - in particular with the literature that this BIP is based on. If there are important differences we should mention it somewhere - if only for documentation.

[Practical Schnorr Threshold Signatures Without the Algebraic Group Model](https://eprint.iacr.org/2023/899.pdf) ("Olaf"):
- **Agreement**: If some honest party outputs true, then every honest party will output true.

The [Secure Multi-Party Computation without Agreement](https://link.springer.com/content/pdf/10.1007/s00145-005-0319-z.pdf) ("GL") solves a problem that is different from `Eq`, namely "broadcast with abort" (one party broadcasts a single value instead of all parties comparing their values as in `Eq`):
- **Agreement**: Byzantine Agreement (All honest parties receive their outputs or all honest parties abort)
- **Broadcast**: Essentially synonym for Byzantine generals (or agreement) problem
- **Broadcast with abort**:  has "agreement for broadcast with abort" (Definition 9)
- **Agreement for broadcast with abort**: If an honest party receives an output, then all honest parties receive the output or abort

Abort in this protocol roughly corresponds to the "in-progress" state of `Eq`.

In the master branch:
- **Consistency**: If some honest participant outputs SUCCESS, no other honest participant outputs FAIL.
- **Conditional Termination**: If some honest participant outputs SUCCESS, then all other participants will (eventually) output SUCCESS.
- **Agreement** is mentioned informally (and the name of the "Ensuring Agreement" section)

From this overview I see the following possible improvements.
1. Conditional Termination implies Consistency, so we could remove the latter.
2. We could rename Conditional Termination to Agreement. The property is equal to Agreement in Olaf and roughly equal to Agreement as defined for Broadcast with Aborts in GL. However, I like Conditional Termination. It's a clearly distinct from Byzantine Agreement.

jonasnick:
This is a suggestion to fix #10. I can't say I'm 100% happy with it, but I feel like this is better than what we have now. I don't see a way around inventing our own definitions as I haven't found the exact properties in the literature.

The entire "Ensuring Agreement" section could be cleaned up (e.g., we start by explaining agreement and why it's important, but we don't mention integrity besides the defintion), but I thought it would be best to agree on the terminology first.


real-or-random:
>  Eq: cond. agreement is not a necessary property
>
>[...] it makes sense to not view it as an essential property since our Eq. protocol before it was certifying only fulfilled consistency. 

It's true that we didn't achieve the property at first, but I think this was at a stage where still figuring things out, and thought we can't get around some last-resort out-of-band mechanism to convince other signers that the protocol finished.

I think if we're honest to ourselves, then the property is essential. If we don't have it, then we may end up in a state where someone has terminated and thus sent coins to the threshold key, but others are stuck. As long as we don't tell users how to get them unstuck, the coins are effectively lost. (And I think, we don't know what to tell them.)  

> Also, conditional agreement implies consistency, so it's a bit confusing to list them both as necessary properties.

Agreed, though I think was left over from some WIP. We could alternatively just remove

jonasnick:
> I think if we're honest to ourselves, then the property is essential.

That's a good point. If a protocol does not satify "Conditional Terminating Agreement" ("If some honest participant outputs True and the delivery of messages between honest participants is guaranteed, then all honest participants output True") then the implication is that there's not even an out-of-band mechanism to get participants unstuck. Hence, I agree that this could lead to a situation where coins can't be recovered.

jonasnick:
I removed "Consistency".

README.md
```diff
@@ -553,17 +553,16 @@ As a consequence, even if Eq appears to be stuck, the caller must not assume (e.
...
 More formally, Eq must fulfill the following properties:
  - Integrity: If some honest participant outputs True, then for every pair of values x and x' input provided by two honest participants, we have x = x'.
- - Consistency: If some honest participant outputs True, no other honest participant outputs False.
- - Conditional Termination: If some honest participant outputs True, then all honest participants will (eventually) output True.
-<!-- The latter two properties together are equivalent to Agreement in the paper. -->
+ - Conditional Agreement: If some honest participant outputs True and the delivery of messages between honest participants is guaranteed, then all honest participants output True.
 
-Optionally, the following property is desired but not always achievable:
- - (Full) Termination: All honest participants will (eventually) output True or False.
+If the protocol has conditional agreement and two honest participants have `x` and `x'` such that `x != x'`, the protocol will not terminate.
```
real-or-random, line 563: s/will not/may not. or phrased differently:
```suggestion
Conditional agreement does *not* guarantee that the protocol terminates if two honest participants have `x` and `x'` such that `x != x'`.
```

real-or-random (commented):
ACK

I think the "eventual" can be improved further, but this requires a separate paragraph about the asynchronous model etc.

README.md
jonasnick: Fixed. The "will not" wording should appear in the context of certifying_eq.

PR #14, opened 2024-01-31
Focus on a single DKG

real-or-random:
Does this make sense roughly? I think it's indeed much simpler. I moved some stuff to "parking lot" for now...


jonasnick (approved):
Makes sense to me. Nice!

PR #19, opened 2024-02-21
remove seed_ from chilldkg

jonasnick:
Hostkeys and threshold are already hashed into the seed used for generating shares in encpedpop.

Maybe there's some value in domain-separating the seed used in EncPedPop vs. ChillDKG such that different shares are generated when the same seed (and same keys and `t`) are used in EncPedPop vs. ChillDKG. We could achieve that with just
```python
seed_ = kdf(seed, "ChillDKG", b'')
```


real-or-random (commented):
ACK

PR #23, opened 2024-03-20
Big name bikeshedding PR 

real-or-random:
Best reviewed commit-by-commit.


reference/encpedpop.py
```diff
@@ -60,18 +60,18 @@ class Broadcast1(NamedTuple):
...
     self_share: Scalar
-    simpl_state: simplpedpop.SignerState1  # TODO Move up?
+    simpl_state: simplpedpop.SignerState  # TODO Move up?
 
 
-def signer_round1(
-    seed: bytes, t: int, n: int, my_deckey: bytes, enckeys: List[bytes], my_idx: int
-) -> Tuple[SignerState1, Unicast1]:
+def signer_round(
+    seed: bytes, t: int, n: int, deckey: bytes, enckeys: List[bytes], idx: int
```
jonasnick, line 73: For clarity, maybe better idx -> participant_idx or signer_idx. I think this makes it significantly clearer what it is.

reference/encpedpop.py
```diff
@@ -60,18 +60,18 @@ class Broadcast1(NamedTuple):
...
+    idx: int
     self_share: Scalar
-    simpl_state: simplpedpop.SignerState1  # TODO Move up?
+    simpl_state: simplpedpop.SignerState  # TODO Move up?
 
 
-def signer_round1(
-    seed: bytes, t: int, n: int, my_deckey: bytes, enckeys: List[bytes], my_idx: int
-) -> Tuple[SignerState1, Unicast1]:
+def signer_round(
```
jonasnick, line 72: signer or participant?

reference/simplpedpop.py
```diff
@@ -39,46 +35,48 @@ def pop_verify(pop: Pop, pubkey: bytes, idx: int):
...
-                for P in self.first_ges + self.remaining_ges
+                for P in self.coms_to_secrets + self.sum_coms_to_nonconst_terms
             ]
         ) + b"".join(self.pops)
 
 
-def aggregate_vss_commitments(
-    first_ges: List[GE], remaining_ges: List[GE], t: int, n: int
+def assemble_sum_vss_commitment(
+    coms_to_secrets: List[GE], sum_coms_to_shares: List[GE], t: int, n: int
```
jonasnick, line 62: sum_coms_to_shares -> sum_coms_to_nonconst_terms?

reference/encpedpop.py
```diff
@@ -80,40 +80,38 @@ def signer_round1(
...
+    self_share = shares[idx]
+    smsg = SignerMsg(simpl_smsg, enc_shares)
+    state = SignerState(t, deckey, enckeys, idx, self_share, simpl_state)
+    return state, smsg
 
 
 def signer_pre_finalize(
-    state1: SignerState1,
-    broad1: Broadcast1,
+    state1: SignerState,
```
jonasnick, line 104: state1 -> state

reference/reference.py
```diff
@@ -193,10 +196,13 @@ async def chilldkg_coordinate(
         simpl_round1_in, enc_shares = await chans.receive_from(i)
         simpl_round1_ins += [simpl_round1_in]
         all_enc_shares_sum = [all_enc_shares_sum[j] + enc_shares[j] for j in range(n)]
-    simpl_round1_outs = simplpedpop.coordinator_round1(simpl_round1_ins, t)
+    simpl_round1_outs = simplpedpop.coordinator_round(simpl_round1_ins, t)
```
jonasnick, line 199: why not drop the 1 in simpl_round1_outs (and ins)?

reference/reference.py
```diff
@@ -262,16 +268,16 @@ def chilldkg_recover(
...
         raise InvalidBackupError("Seed and backup don't match") from e
 
     shares_sum = encpedpop.decrypt_sum(
-        all_enc_shares_sum[my_idx], my_hostseckey, hostpubkeys, my_idx, enc_context
+        all_enc_shares_sum[idx], hostseckey, hostpubkeys, idx, enc_context
     )
     # TODO: don't call full round1 function
-    (state1, (_, _)) = encpedpop.signer_round1(
-        seed, t, len(hostpubkeys), my_hostseckey, hostpubkeys, my_idx
+    (state1, (_, _)) = encpedpop.signer_round(
```
jonasnick, line 279: state1 -> state?

reference/tests.py
```diff
@@ -71,18 +71,16 @@ def simulate_encpedpop(seeds, t):
 
     enckeys = [out[1] for out in round0_outputs]
     for i in range(n):
-        my_deckey = round0_outputs[i][0]
-        round1_outputs += [
-            encpedpop.signer_round1(seeds[i], t, n, my_deckey, enckeys, i)
-        ]
+        deckey = round0_outputs[i][0]
+        round1_outputs += [encpedpop.signer_round(seeds[i], t, n, deckey, enckeys, i)]
```
jonasnick, line 75: round1_outputs -> round_outputs (same for simplpedpop)?

reference/reference.py
```diff
@@ -42,21 +42,21 @@ def chilldkg_session_params(
...
+) -> Tuple[ChillDKGStateR1, simplpedpop.SignerMsg, List[Scalar]]:
+    hostseckey, hostpubkey = chilldkg_hostkey_gen(seed)
     (hostpubkeys, t, params_id) = params
     n = len(hostpubkeys)
 
-    my_idx = hostpubkeys.index(my_hostpubkey)
-    enc_state1, (vss_commitment_ext, enc_gen_shares) = encpedpop.signer_round1(
-        seed, t, n, my_hostseckey, hostpubkeys, my_idx
+    idx = hostpubkeys.index(hostpubkey)
+    enc_state1, (vss_commitment_ext, enc_gen_shares) = encpedpop.signer_round(
```
jonasnick, line 56: enc_state1 -> enc_state

reference/reference.py
```diff
@@ -127,15 +128,17 @@ async def chilldkg(
...
+    # TODO We should probably not just return None here but raise instead.
+    # Raising a specific exception is also better for testing.
     if dkg_output is None:
         return None
 
     return (dkg_output, chilldkg_backup(state2, cert))
 
 
-def certifying_eq_round1(my_hostseckey: bytes, x: bytes) -> bytes:
+def certifying_eq_round1(hostseckey: bytes, x: bytes) -> bytes:
```
jonasnick, line 139: should be certifying_eq_round?

jonasnick: and similar, eq_round in chilldkg

reference/encpedpop.py
jonasnick, line 72: maybe something like `_round_start` makes more sense since this function doesn't really do a full communication round. Then `_round_reply` for the coordinator.

reference/simplpedpop.py
```diff
@@ -93,72 +91,73 @@ class DKGOutput(NamedTuple):
...
             raise InvalidContributionError(i, "Participant sent invalid commitment")
-        else:
-            if not pop_verify(pops[i], first_ges[i].to_bytes_xonly(), i):
-                raise InvalidContributionError(
-                    i, "Participant sent invalid proof-of-knowledge"
-                )
-    vss_commitment = aggregate_vss_commitments(first_ges, remaining_ges, t, n)
-    if not vss_commitment.verify(my_idx, shares_sum):
+        # This can be optimized: We serialize the coms_to_secrets[i] here, but
+        # pop_verify/schnorr_veriy will need to deserialize it again, which
```
jonasnick, line 151: typo: veriy

jonasnick (commented):
Concept ACK to most

The README (chilldkg function signatures) and diagram needs an update as well.

jonasnick (commented):
Other potential renames include:
- shared pubkey -> group pubkey (matches FROST RFC draft). I don't have a strong opinion on this. Maybe very light preference for "shared".
- signer pubkey:
    - doesn't have a short name in the FROST RFC. At one point it's called `PK_i: the public key for the i-th participant`.
    - is called `pubshare` in Jesse/Siv's signing BIP
    - is called "individual pubkey" in BIP MuSig
-  share:
    - called secshare in Jesse/Siv's signing BIP
    - "Signer secret key share" `sk_i` in the FROST RFC

reference/encpedpop.py
real-or-random: Happy to switch to "participant", but that should go to a separate PR (or just commit).

reference/reference.py
real-or-random: Yep, I haven't touched only some parts in reference.py so far, i.e., those that were necessary to make the code work or those that were found by search and replace. I'll make reference.py consistent when I go over chilldkg. The point of this PR is more to agree on the terms.

reference/encpedpop.py
real-or-random: Yeah, I'm not happy with "round" anyway for these functions, because that term is a communication thing, but the functions represent pure computations. I had considered "msg", but that's not great either because the functions also compute the state and not only the message to be sent.

 Now that I think about it again, I really like "step". The functions really represent a protocol step (or computation step) of a party.  What do you think?

real-or-random:
> Other potential renames include:
> 
>     * shared pubkey -> group pubkey (matches FROST RFC draft). I don't have a strong opinion on this. Maybe very light preference for "shared".

The advantage of "shared" is that it expresses that everyone has a share of it. But that's also the disadvantage: It can be confused with shares from VSS etc, because we have so many sharings (additive sharing, Shamir sharing, Shamir sharing of the final secret etc...) "group pubkey" is not bad. What about "threshold pubkey"? That's similar in spirit, I think.

  We could also use "aggregate", but I'd rather reserve this for multisignatures with key aggregation.  

> 
>     * signer pubkey:
>       
>       * doesn't have a short name in the FROST RFC. At one point it's called `PK_i: the public key for the i-th participant`.
>       * is called `pubshare` in Jesse/Siv's signing BIP
>       * is called "individual pubkey" in BIP MuSig
>     * share:
>       
>       * called secshare in Jesse/Siv's signing BIP
>       * "Signer secret key share" `sk_i` in the FROST RFC


I prefer "secret key"/"public key" over "secret share"/"public share". It's simply more common and then also easier to handle  implementors? 

(If we don't use "share" here, this will also be an argument against "shared" pubkey).

And "Individual" is a good addition in case there's risk of confusion with the shared/group/threshold pubkey. Being consistent with BIP MuSig is a good idea.

nickfarrow:
We currently have `joint public key` (referred as this only once in FROST paper) but I like `group public key` or `group public key` more. 
Perhaps one minor reason to favour `group` is that the word "share" is absolutely everywhere (at least in our code).

+1 for `secret share` or `secret key share`.

real-or-random:
> We currently have `joint public key` (referred as this only once in FROST paper) but I like `group public key` or `shared public key` more. Perhaps one minor reason to favour `group` is that the word "share" is absolutely everywhere (at least in our code).

Hehe, "joint public key" is also nice. What do you think about "threshold public key"?


---
> except I do prefer `secret share` over `secret key` since they are actual shares!

Yeah, indeed. I think my preference for "secret key" is that people will know that this is what they can use for signing.  But that can be a disadvantage. If the only purpose of this thing is to participate in a FROST session for this group, then "share" may be a good idea, to clarify that you should not use this thing like a private key. Now that I'm writing this, I like "share".

But we should probably be consistent with the public part, i.e., we should take either the pair secret share/public share, or the pair secret key/(individual) public key.

real-or-random:
Just had a call with @jonasnick. We settled on the following terms for now.
 - "participant"
   - it's a DKG, not a signature scheme
   - the output of the DKG is currently to be used with FROST, but this may change in the future 
 - "secret share/public share" 
   - "individual" is then implied and optional
   - this also differentiates them from host secret keys / host public keys
   - this emphasizes that you should not single-sign using these secrets
 -  threshold public key
    - we think it's natural in this setting  
    - "group public key" is also okay, but "threshold" is a bit more precise
  
I'll probably implement this in a later PR.

reference/encpedpop.py
real-or-random: will do when renaming signer to participant

real-or-random:
Fixed and marked-resolved some of these. I'll take care of the rest when going over the chilldkg code or when doing further renames.

PR #42, opened 2024-08-14
Replace ChillDKG seed with hostseckey

### LINKED ISSUE #28: Should we make the hostseckey the cryptographic seed?
Currently, we derive everything from the seed, including the hostseckey. We could instead let the user pass the hostseckey and then, if we need randomness for the VSS, derive it from the hostseckey.

Advantages:
 - This fits BIP32 better, which produces "keys" (and not seeds). The BIP32 derived key pair will just be the host key pair.
 - Fewer derivations. 

Disadvantage:
  - This fits BIP32 better. People may start doing public derivation, or receive coins under their BIP32 host public key. (But wallets already know how to deal with this in other multisig scenarios, e.g., wallets use different derivation paths for single sig and multisig)

jonasnick:
WIP because the markdown is inconsistent and confusing. For example, the "host secret key" is also called "long-term secret key", "host secret key", "device secret key" or just "secret key".

Fixes #28


python/chilldkg_ref/util.py
```diff
@@ -18,6 +18,10 @@ class SeedError(ValueError):
     pass
 
 
+class SecretKeyError(ValueError):
+    pass
+
+
```
real-or-random, line 24: Can we remove `SeedError` then?

README.md
```diff
@@ -458,7 +457,7 @@ If there is no dedicated coordinator, one of the participants can act as the coo
 
 ### Inputs and Output
 
-The inputs of a session consist of a long-term *secret seed* (individual to each participant, not provided by the coordinator) and public *session parameters* (common to all participants and the coordinator).
+The inputs of a session consist of a long-term *secret key* (individual to each participant, not provided by the coordinator) and public *session parameters* (common to all participants and the coordinator).
```
real-or-random, line 460:
```suggestion
The inputs of a session consist of a long-term *host secret key* (individual to each participant, not provided by the coordinator) and public *session parameters* (common to all participants and the coordinator).
```
(here, "long-term" is a useful description when introducing the term)

README.md
```diff
@@ -111,12 +111,12 @@ This is sufficient to exclude the catastrophic failure described in the previous
...
 <!-- TODO Call this restore instead of recovery? -->
 As an additional feature of ChillDKG, the DKG outputs for any signing device can be fully recovered from
-a backup of a single secret per-device seed,
+a backup of a single secret per-device secret key,
 (the essential parts of) the public transcripts of the DKG sessions,
 and the corresponding success certificates.
 To simplify the interface, we combine the transcript data and the session certificate into a single byte string called the *recovery data*,
 which is common to all participants and does not need to be kept confidential.
-Recovering a device that has participated in a DKG session then requires just the device seed and the recovery data,
+Recovering a device that has participated in a DKG session then requires just the device's secret key and the recovery data,
```
real-or-random, line 119:
```suggestion
Recovering a device that has participated in a DKG session then requires just the device's host secret key and the recovery data,
```

README.md
```diff
@@ -111,12 +111,12 @@ This is sufficient to exclude the catastrophic failure described in the previous
 
 <!-- TODO Call this restore instead of recovery? -->
 As an additional feature of ChillDKG, the DKG outputs for any signing device can be fully recovered from
-a backup of a single secret per-device seed,
+a backup of a single secret per-device secret key,
```
real-or-random, line 114:
```suggestion
a backup of a single *host secret key* specific to the device.
```

I think "device" is fine here. More technically, it's the participant, but at the level of abstraction of this paragraph, device is more tangible.

README.md
```diff
@@ -129,7 +129,7 @@ In summary, we aim for the following design goals:
  - **Conditional agreement**: If a ChillDKG session succeeds for one honest participant, this participant will be able to convince every other honest participant that the session has succeeded.
  - **No restriction on threshold**:  Like the FROST signing protocol, ChillDKG supports any threshold `t <= n`, including `t > n/2` (also called "dishonest majority").
  - **Broad applicability**:  ChillDKG supports a wide range of scenarios, from those where the signing devices are owned and connected by a single individual to those where multiple owners manage the devices from distinct locations.
- - **Simple backups**: The capability of ChillDKG to recover devices from a static seed and public recovery data avoids the need for secret per-session backups, enhancing user experience.
+ - **Simple backups**: The capability of ChillDKG to recover devices from a static secret key and public recovery data avoids the need for secret per-session backups, enhancing user experience.
```
real-or-random, line 132:
```suggestion
 - **Simple backups**: The capability of ChillDKG to recover devices from the host secret key together with public recovery data avoids the need for secret per-session backups, enhancing user experience.
```

README.md
```diff
@@ -407,15 +407,14 @@ e.g., during a request to participate in a FROST signing session.
 ChillDKG constructs a transcript `eq_input` by appending to the transcript of EncPedPop the vector `enc_secshare`.
 This ensures that all participants agree on all encrypted shares,
 and as a consequence,
-the entire DKG output of a successful ChillDKG participant can be deterministically reproduced from a secret per-participant seed and the transcript.
+the entire DKG output of a successful ChillDKG participant can be deterministically reproduced from a per-participant secret key and the transcript.
```
real-or-random, line 410:
```suggestion
the entire DKG output of a successful ChillDKG participant can be deterministically reproduced from a per-participant *host secret key* and the transcript.
```

README.md
```diff
@@ -407,15 +407,14 @@ e.g., during a request to participate in a FROST signing session.
...
 
 This property is leveraged to offer a backup and recovery functionality:
 ChillDKG outputs a string called *recovery data* which is the concatenation of the transcript `eq_input` and the success certificate `cert`.
-The recovery data, which is the same for every participant, can be used by any participant together with the seed to recover the full output of the DKG session.
+The recovery data, which is the same for every participant, can be used by any participant together with the secret key to recover the full output of the DKG session.
```
real-or-random, line 414:
```suggestion
The recovery data, which is the same for every participant, can be used by any participant together with the host secret key to recover the full output of the DKG session.
```

README.md
```diff
@@ -473,8 +472,8 @@ unlike deterministically derived secret keys [[BIP32](https://github.com/bitcoin
 not be rederived solely from the participant's seed.
 
 To facilitate backups of a DKG session,
-ChillDKG offers the possibility to recover a participant's DKG output from the participant's seed and the recovery data of the specific session,
-As a result, a full backup of a participant consists of the seed as well as the recovery data of all DKG sessions the participant has successfully participated in.
+ChillDKG offers the possibility to recover a participant's DKG output from the participant's secret key and the recovery data of the specific session,
+As a result, a full backup of a participant consists of the secret key as well as the recovery data of all DKG sessions the participant has successfully participated in.
```
real-or-random, line 476:
```suggestion
ChillDKG offers the possibility to recover a participant's DKG output from the participant's host secret key and the recovery data of the specific session,
As a result, a full backup of a participant consists of the host secret key as well as the recovery data of all DKG sessions the participant has successfully participated in.
```

README.md
```diff
@@ -483,15 +482,15 @@ Moreover, the recovery data contains secrets only in encrypted form and is self-
 so that it can, in principle, be stored with an untrusted third-party backup provider.
 Users should, however, be aware that the session parameters (the threshold and the host public keys) and public parts of the DKG output (the threshold public key and the public shares) can be inferred from the recovery data, which may constitute a privacy issue.
 
-Keeping seed backups accessible and secure is hard (typically similarly hard as keeping the participant devices themselves).
-As a consequence, it may not be an unreasonable strategy in a threshold setup not to perform backups of seeds at all,
+Keeping secret key backups accessible and secure is hard (typically similarly hard as keeping the participant devices themselves).
```
real-or-random, line 485:
```suggestion
Keeping backups of the secret key accessible and secure is hard (typically similarly hard as keeping the participant devices themselves).
```

python/chilldkg_ref/chilldkg.py
```diff
@@ -124,44 +125,34 @@ def certeq_coordinator_step(sigs: List[bytes]) -> bytes:
 ###
 
 
-def hostseckey(seed: bytes) -> bytes:
-    return prf(seed, "chilldkg hostseckey")
+def hostpubkey_gen(hostseckey: bytes) -> bytes:
```
real-or-random, line 128:
```suggestion
def hostpubkey(hostseckey: bytes) -> bytes:
```
I like this a bit more. 
 - I don't think the public key is "generated" at this point. It's rather that the entire key pair is generated when the hostseckey is created.
 - For consistency with the other names.  We also don't use `certeq_message_gen` or `params_id_gen`.

real-or-random (commented):
Concept ACK for the reasons explained in #28.

perhaps the `seed` argument in encpedpop `participant_step1` should then also be renamed to `deckey`  (because that's what we pass anyway)

> WIP because the markdown is inconsistent and confusing. For example, the "host
secret key" is also called "long-term secret key", "host secret key", "device
secret key" or just "secret key".

Urghs yeah. What about calling it "host secret key" always, but adding descriptive prefixes if necessary? See my inline suggestions.

README.md
jonasnick: done

README.md
jonasnick: done

README.md
jonasnick: Agree, done

README.md
jonasnick: done

README.md
jonasnick: done

README.md
jonasnick: done

python/chilldkg_ref/chilldkg.py
jonasnick: done. I renamed the local conflicting variables to `hostpk`.

jonasnick:
> perhaps the seed argument in encpedpop participant_step1 should then also be renamed to deckey (because that's what we pass anyway)

`seed` seems to be more correct in the context of encpedpop because the argument is in principle unrelated to the deckey.

real-or-random:
Forgot to push your changes? ;)

real-or-random:
I reverted the naming change. (I think I had a point there, but I really dislike the naming conflict with the local variables...)

Manually pushed to master here as 4617da4a06064cdef44b86a566a42c6ddd42ad67

PR #43, opened 2024-08-14
Encrypt & send "self share" to avoid rederiving in recovery

jonasnick:
Depends on #42

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -16,7 +16,7 @@ from secp256k1proto.bip340 import schnorr_sign, schnorr_verify
 from secp256k1proto.keys import pubkey_gen_plain
 from secp256k1proto.util import int_from_bytes, bytes_from_int
 
-from .vss import VSS, VSSCommitment
+from .vss import VSSCommitment
 from . import encpedpop
 from .util import (
     BIP_TAG,
@@ -639,9 +639,6 @@ def recover(
 
         # Decrypt share
         enc_context = encpedpop.serialize_enc_context(t, hostpubkeys)
-        session_seed = encpedpop.derive_session_seed(
-            hostseckey, pubnonces[idx], enc_context
-        )
         secshare = encpedpop.decrypt_sum(
             hostseckey,
             hostpubkeys[idx],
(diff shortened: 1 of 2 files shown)


real-or-random (approved):
utACK mod the rebase

Independently of making the randomness source flexible, this gets rid of the somewhat ugly index handling. That was supposed to save 32 bytes in the recovery data, and I think size is important for recovery data (e.g., think of recovering it from a sheet of paper), but those 32 bytes are probably not worth it. The size of the recovery data for a 2-of-3 is 556 bytes after this PR. That's an order of magnitude larger than what we could save, so it's not worth the effort. (Even with musig2 cert, the size is 428 bytes after this PR).

real-or-random:
Rebased this, but I also pushed other stuff to master, so this needs rebase again, I can take care of this....

Now that I think about this again, it's not just the 32 bytes in communication, but also the ECDH computation.... 

Here's a middle way: We encrypt to ourselves, but by deriving a `self_pad` from the decryption key using a hash function without ECDH. This means one can easily recover with the decryption key and without the need to follow our exact derivation method deep down to share generation in SimplPedPod. IIUC, this is what @LLFourn asked for (correct me if I'm wrong!), but we still save the ECDH computation, just not the 32 bytes on the wire. 

A downside is that the logic is slightly more complicated than what is proposed here. But it should still be simpler than the index stuff that we have right now: There's no need to deal with indices because we can simply compare enckeys to check if we're encryption to ourselves/decryption from ourselves. I'll work on a PR as an alternative to this one.

jonasnick:
Sounds like a reasonable improvement. ECDH to self does not make a lot of sense. Depending on how complex the implementation of this special case is, maybe it's overkill because we're only saving 1 ecmult in step1 which (off the top of my head) has t ecmult_gens (VSS) + n ecmults (encryption) + 1 ecmult (signature) (and in chilldkg we have 1 more ecmult_gen to compute the hostpubkey). Concept ACK if the implementation is simple.

real-or-random:
Rebased again, so this is in principle ready for merge, but I prefer #51.

real-or-random:
Superseded by merged #51.

PR #46, opened 2024-09-02
fix `sum_coms_to_nonconst_terms` formula

siv2r:
I believe there's an error in the current formula. It appears to sum the $$t-1$$ polynomial coefficients of the same participant, instead of performing a component-wise sum across all participants.


jonasnick (approved):
Yes, thanks @siv2r.

real-or-random (commented):
post-merge ACK

thanks!

PR #48, opened 2024-10-01
encpedpop: Unify deckey and seed

real-or-random:
ChillDKG anyway passes the hostseckey for both, so this was potentially confusing to readers of the code. This change is in line with #42.


jonasnick (commented):
Hm, I thought this was more confusing to readers because the change implies that what was formerly the `seed` in step1 has to be identical to the `deckey` in step2, but they are entirely unrelated. But I have no strong opinion either way. If we go ahead with the change we should also update the naming in the encpedpop tests.

real-or-random:
Agreed, closing in favor of #49.

Issue #50, opened 2024-10-02
prf vs tagged_hash_bip_dkg


real-or-random:
The difference between these functions is perhaps a bit mood, or at least academic, in particular after #42.  Consider simplifying them into a single function.

jonasnick:
Agree, suggesting to remove prf, see #54

PR #51, opened 2024-10-02
Encrypt & send "self share" to avoid rederiving in recovery (with optimized ECDH)

real-or-random:
Alternative to #43.

As opposed to what I claimed in #43, I've stuck to comparing the indices instead of the enckeys, because then don't need the array of all enckeys during decryption. Moreover, it seems a bit cleaner from a cryptographic point of view: We know for sure that our index belongs only to us, whereas our enckey could also belong to someone else in the sense that the attacker can claim to have the same enckey (at least in pure EncPedPop where we don't disallow duplicate pubkeys).

I doubt that comparing indices instead of enckeys makes a difference in terms of code complexity. The ugly thing (also for implementers) on master is the omission of the self_share in the array on the wire, which is hard to get right (easy off-by-one errors etc), and which this PR also gets rid off.

### DIFF
--- a/python/chilldkg_ref/encpedpop.py
+++ b/python/chilldkg_ref/encpedpop.py
@@ -28,45 +28,58 @@ def ecdh(
     return Scalar(int_from_bytes(tagged_hash_bip_dkg("encpedpop ecdh", data)))
 
 
+def self_pad(deckey: bytes, context_: bytes) -> Scalar:
+    return Scalar(
+        int_from_bytes(
+            prf(seed=deckey, tag="encaps_multi self_pad", extra_input=context_)
+        )
+    )
+
+
 def encaps_multi(
-    secnonce: bytes, pubnonce: bytes, enckeys: List[bytes], context: bytes, idx: int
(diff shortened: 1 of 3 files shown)


python/chilldkg_ref/encpedpop.py
```diff
@@ -28,45 +28,58 @@ def ecdh(
     return Scalar(int_from_bytes(tagged_hash_bip_dkg("encpedpop ecdh", data)))
 
 
+def self_pad(deckey: bytes, context_: bytes) -> Scalar:
+    return Scalar(
+        int_from_bytes(
+            prf(seed=deckey, tag="encaps_multi self_pad", extra_input=context_)
```
real-or-random, line 34: We could also make the derivation even explicit here, to make sure it's easier to reimplement (e.g, by getting rid of `prf` entirely, see #50).

jonasnick:
Needs rebase

jonasnick (approved):
Implementation looks quite straightforward. ACK.

Issue #37, opened 2024-07-10
Implement Identifiable Aborts (take 2)


real-or-random:
Implement the idea from https://github.com/BlockstreamResearch/bip-frost-dkg/issues/9#issuecomment-2220091151, ideally in a PR so we can judge the complexity.

Steps:
 - [x] ~Coordinator should probably check the messages received as far as possible (https://github.com/BlockstreamResearch/bip-frost-dkg/issues/29) and raise `InvalidContributionExcpetion`~
 - [x] The actual implementation of the idea (#52, #75)
 - [x] Document exceptions properly (#52, #62, #65)
 - [x] Document `blame` functions (#72, #75)
 - [x] Change BIP text to explain IA (this needs to say that the coordinator is fully trusted for IA) (#71, #74)

real-or-random:
One natural way to do this is to let each coordinator_step return additional blame data, separate for each participant.  This could be an easy way of making IA optional: Just skip computing and sending this data around.

Not sure if it's worth making IA optional, but at least the additional complexity over non-optional IA is tiny.

real-or-random:
> One natural way to do this is to let each coordinator_step return additional blame data, separate for each participant. This could be an easy way of making IA optional: Just skip computing and sending this data around.

Having separate blame data also has the advantage that "main" CoordinatorMsg remains the same for everyone. That is, if you don't need IA, you can get away with the coordinator sending the exact same bytes to everyone (without wasting communication), which may be simpler.

PR #52, opened 2024-10-17
Identifiable aborts

real-or-random:
Solves partly #37, namely the actual implementation of the idea. 

Draft because there are some minor cleanups left, see added FIXMEs.

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -21,9 +21,11 @@ from . import encpedpop
 from .util import (
     BIP_TAG,
     tagged_hash_bip_dkg,
+    ProtocolError,
     SecretKeyError,
     ThresholdError,
-    InvalidContributionError,
+    FaultyParticipantOrCoordinatorError,
+    FaultyCoordinatorError,
 )
 
 __all__ = [
@@ -39,7 +41,8 @@ __all__ = [
     # Exceptions
     "SecretKeyError",
     "ThresholdError",
-    "InvalidContributionError",
+    "FaultyParticipantOrCoordinatorError",
+    "FaultyCoordinatorError",
     "InvalidRecoveryDataError",
     "DuplicateHostpubkeyError",
     "SessionNotFinalizedError",
@@ -62,15 +65,15 @@ __all__ = [
 ###
 
 
-class DuplicateHostpubkeyError(ValueError):
+class DuplicateHostpubkeyError(ProtocolError):
     pass
 
 
-class SessionNotFinalizedError(Exception):
+class SessionNotFinalizedError(ProtocolError):
     pass
 
 
-class InvalidRecoveryDataError(Exception):
+class InvalidRecoveryDataError(ValueError):
     pass
 
 
@@ -219,7 +222,7 @@ def params_validate(params: SessionParams) -> None:
         try:
             _ = GE.from_bytes_compressed(hostpubkey)
         except ValueError as e:
-            raise InvalidContributionError(
+            raise FaultyParticipantOrCoordinatorError(
                 i, "Participant has provided an invalid host public key"
             ) from e
 
@@ -228,7 +231,7 @@ def params_validate(params: SessionParams) -> None:
 
 
 def params_id(params: SessionParams) -> bytes:
-    """Returns the parameters ID, a unique representation of the`SessionParams`.
+    """Return the parameters ID, a unique representation of the `SessionParams`.
 
     In the common scenario that the participants obtain host public keys from
     the other participants over channels that do not provide end-to-end
@@ -246,8 +249,8 @@ def params_id(params: SessionParams) -> bytes:
         bytes: The parameters ID, a 32-byte string.
 
     Raises:
-        InvalidContributionError: If `hostpubkeys[i]` is not a valid public key
-            for some `i`, which is indicated as part of the exception.
+        FaultyParticipantOrCoordinatorError: If `hostpubkeys[i]` is not a valid
+            public key for some `i`, which is indicated in the exception.
         DuplicateHostpubkeyError: If `hostpubkeys` contains duplicates.
         ThresholdError: If `1 <= t <= len(hostpubkeys)` does not hold.
         OverflowError: If `t >= 2^32` (so `t` cannot be serialized in 4 bytes).
@@ -395,8 +398,8 @@ def participant_step1(
         ValueError: If the participant's host public key is not in argument
         `hostpubkeys`.
         SecretKeyError: If the length of `hostseckey` is not 32 bytes.
-        InvalidContributionError: If `hostpubkeys[i]` is not a valid public key
-            for some `i`, which is indicated as part of the exception.
+        FaultyParticipantOrCoordinatorError: If `hostpubkeys[i]` is not a valid
+            public key for some `i`, which is indicated in the exception.
         DuplicateHostpubkeyError: If `hostpubkeys` contains duplicates.
         ThresholdError: If `1 <= t <= len(hostpubkeys)` does not hold.
         OverflowError: If `t >= 2^32` (so `t` cannot be serialized in 4 bytes).
@@ -445,9 +448,9 @@ def participant_step2(
 
     Raises:
         SecKeyError: If the length of `hostseckey` is not 32 bytes.
-        InvalidContributionError: If `cmsg1` is invalid. This can happen if
-            another participant has sent an invalid message to the coordinator,
-            or if the coordinator has sent an invalid `cmsg1`.
+        FaultyParticipantOrCoordinatorError: If `cmsg1` is invalid. This can
+            happen if another participant has sent an invalid message to the
+            coordinator, or if the coordinator has sent an invalid `cmsg1`.
 
             Further information is provided as part of the exception, including
             a hint about which party might be to blame for the problem. The hint
@@ -543,8 +546,8 @@ def coordinator_step1(
             `coordinator_finalize` call).
 
     Raises:
-        InvalidContributionError: If `hostpubkeys[i]` is not a valid public key
-            for some `i`, which is indicated as part of the exception.
+        FaultyParticipantOrCoordinatorError: If `hostpubkeys[i]` is not a valid
+            public key for some `i`, which is indicated in the exception.
         DuplicateHostpubkeyError: If `hostpubkeys` contains duplicates.
         ThresholdError: If `1 <= t <= len(hostpubkeys)` does not hold.
         OverflowError: If `t >= 2^32` (so `t` cannot be serialized in 4 bytes).
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -9,7 +9,7 @@ their arguments and return values, and the exceptions they raise; see also the
 """
 
 from secrets import token_bytes as random_bytes
-from typing import Tuple, List, NamedTuple, NewType, Optional
+from typing import Tuple, List, NamedTuple, NewType, Optional, NoReturn
 
 from secp256k1proto.secp256k1 import Scalar, GE
 from secp256k1proto.bip340 import schnorr_sign, schnorr_verify
@@ -26,6 +26,7 @@ from .util import (
     ThresholdError,
     FaultyParticipantOrCoordinatorError,
     FaultyCoordinatorError,
+    UnknownFaultyPartyError,
 )
 
 __all__ = [
@@ -43,6 +44,7 @@ __all__ = [
     "ThresholdError",
     "FaultyParticipantOrCoordinatorError",
     "FaultyCoordinatorError",
+    "UnknownFaultyPartyError",
     "InvalidRecoveryDataError",
     "DuplicateHostpubkeyError",
     "SessionNotFinalizedError",
@@ -51,6 +53,7 @@ __all__ = [
     "DKGOutput",
     "ParticipantMsg1",
     "ParticipantMsg2",
+    "CoordinatorBlameMsg",
     "ParticipantState1",
     "ParticipantState2",
     "CoordinatorMsg1",
@@ -256,7 +259,7 @@ def params_id(params: SessionParams) -> bytes:
         OverflowError: If `t >= 2^32` (so `t` cannot be serialized in 4 bytes).
     """
     params_validate(params)
-    (hostpubkeys, t) = params
+    hostpubkeys, t = params
 
     t_bytes = t.to_bytes(4, byteorder="big")  # OverflowError if t >= 2**32
     params_id = tagged_hash_bip_dkg(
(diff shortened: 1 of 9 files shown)


real-or-random:
Okay, I reworked the entire thing and the API is totally different now from the first version. Instead of adding `blame: bool` and similar parameters in the necessary locations, this now adds separate `_blame` functions. In particular, there are `_participant_blame` functions, which can be used after `participant_step2` has failed. Fortunately, this works without the need to redo expensive computations in `participant_blame` (or carry over cached values from `participant_step2` to `participant_blame` via the raised exception, which would be rather inelegant). I think this is much better:
  - The code is simpler and easier to read.
  - No more boolean flags and `Optional`, which are difficult in static typing.
  - People can easily ignore the blame stuff by just not looking at the `_blame` functions.

Still to do:
 - [ ] Add docstrings for the new functions and change docs for the modified exceptions:
 - [ ] Add blaming to the example, i.e., add the code to catch the error in step2 and perform the blame phase so that we have a full interactive ("async") application example with blame. (The example currently runs fine without blame, but this is just because everyone is honest.)
 -  [ ] (The other items in #37.)

Please let me know if this API looks good, then we merge this and I can work on follow-up PRs.

python/chilldkg_ref/simplpedpop.py
```diff
@@ -56,6 +72,10 @@ def to_bytes(self) -> bytes:
         ) + b"".join(self.pops)
 
 
+class BlameRecord(NamedTuple):
+    partial_pubshares: List[GE]
```
jonasnick, line 76: BlameRecord is only created by the coordinator and sent to the participants. Have you considered calling it `CoordinatorBlameMsg` or something? Might make it more clear how it is used.

python/chilldkg_ref/encpedpop.py
```diff
@@ -126,6 +145,11 @@ class CoordinatorMsg(NamedTuple):
     pubnonces: List[bytes]
 
 
+class BlameRecord(NamedTuple):
+    enc_partial_secshares: List[Scalar]
```
jonasnick, line 149: nit: maybe instead of `partial_secshares` we could call them `secshare_terms`, `secshare_addends` or `secshare_summands`. Maybe that'd make it more clear what they are. Similar for `pubshares`.

python/chilldkg_ref/simplpedpop.py
```diff
@@ -161,6 +212,45 @@ def participant_step2(
     return dkg_output, eq_input
 
 
+def participant_blame(
+    state: ParticipantState,
+    secshare: Scalar,
+    partial_secshares: List[Scalar],
+    blame_rec: BlameRecord,
+) -> NoReturn:
```
jonasnick, line 220: Have you considered outputting a separate state in participant_state2 that could be used here? As far as I can see this would
1. Allow `participant_blame` to check explicitly whether `pubshare = sum(partial_pubshares)` and therefore allow not raising an exception if participant_step2 succeeded.
2. In `encpedpop` if the pads were stored in that new state, then they wouldn't have to be recomputed during blaming. You say that "Fortunately, this works without the need to redo expensive computations in participant_blame", but recomputing the pads is a rather expensive computation.

jonasnick (commented):
Looks very good and clear overall. The API makes sense. Left a few comments.

> [ ] Add blaming to the example

Could also output the size of the blaming data, similar to how we output the size of the recovery data.

python/chilldkg_ref/simplpedpop.py
```diff
@@ -104,14 +132,34 @@ def participant_step1(
...
+# other aggregation steps.
+#
+# If EncPedPop is used instead (as a wrapper of SimplPedPop), the coordinator
+# can securely aggregate the encrypted partial secshares into an encrypted
+# secshare by exploiting the additively homomorphic property of the encryption.
+def participant_step2_prepare_secshare(
+    partial_secshares: List[Scalar],
+) -> Scalar:
+    secshare: Scalar = Scalar.sum(*partial_secshares)
+    return secshare
```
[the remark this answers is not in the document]
real-or-random: mypy complains otherwise. Fixed by adding an annotation in secp256k1proto instead.

python/chilldkg_ref/simplpedpop.py
```diff
@@ -122,36 +170,39 @@ def participant_step2(
...
-        # TODO What to raise here? We don't know who was wrong.
-        raise InvalidContributionError(None, "Received invalid secshare.")
+    pubshare = sum_coms.pubshare(idx)
+
+    if not VSSCommitment.verify_secshare(secshare, pubshare):
+        raise UnknownFaultyPartyError(
+            "Received invalid secshare, consider blaming to determine faulty party",
+        )
 
+    pubshares = [sum_coms.pubshare(i) if i != idx else pubshare for i in range(n)]
```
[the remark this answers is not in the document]
real-or-random: I'm not sure if I understand the concern.

> Because we want to save computation of all the the pubshares in case `verify_secshare` fails?

Are you asking why we compute `pubshares` only when if `verify_secshare` fails? That's indeed because we don't need it to call `verify_secshare`. Not sure if this deserves a comment.

python/chilldkg_ref/chilldkg.py
```diff
@@ -511,11 +523,28 @@ def participant_finalize(
...
+) -> Tuple[DKGOutput, RecoveryData]:
+    """Perform a participant's blame step of a ChillDKG session. TODO"""
+    _, idx, enc_state = state1
+    return encpedpop.participant_blame(
+        state=enc_state,
+        deckey=hostseckey,
+        cmsg=cmsg1.enc_cmsg,
+        enc_secshare=cmsg1.enc_secshares[idx],
+        blame_rec=blame_rec.enc_blame_rec,
+    )
```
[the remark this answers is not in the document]
real-or-random: Good catch, fixed.

> Strange that mypy doesn't catch this.

mypy's analysis is sound at least. `encpedpop.participant_blame` is guaranteed not to return, so `participant_blame` will never return anything else than `Tuple[DKGOutput, RecoveryData]`.

python/chilldkg_ref/util.py
```diff
@@ -22,7 +20,19 @@ class ThresholdError(ValueError):
     pass
 
 
-class InvalidContributionError(Exception):
-    def __init__(self, participant: Optional[int], error: str) -> None:
+class ProtocolError(Exception):
+    pass
+
+
+class FaultyParticipantError(ProtocolError):
```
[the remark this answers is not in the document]
real-or-random: Sounds good to me. Explicit is better than implicit. Changed it to `FaultyParticipantOrCoordinatorError`. We could also call it `FaultyPartyError` instead to avoid the long name, but I prefer the explicitness given that this is reference code.  

I may want to improve further this by adding a subclass InvalidHostpubkeyError for the common case when a function gets an invalid hostpubkey.

python/chilldkg_ref/simplpedpop.py
real-or-random: Yes, that's clearer. Done.

python/chilldkg_ref/encpedpop.py
real-or-random: Hm, yeah, there's no obvious great term here. My reason to call this "partial" was the analogy to "partial signature" and "partial verification", where partial means "from a single participant". It's not perfect, but I don't think any of your suggestions are clearly better. Let me keep this unless you feel strongly about it.

python/chilldkg_ref/simplpedpop.py
real-or-random: Agreed. I had overlooked that computing the pads is expensive. Will overhaul this as the next step. We'll need to "return" the state as part of the exception then, but that's fine.

python/chilldkg_ref/util.py
```diff
@@ -22,7 +22,38 @@ class ThresholdError(ValueError):
...
+    This is raised when another participant appears to have sent an invalid
+    protocol message. This error does not necessarily imply that the suspected
+    participant is faulty. It is always possible that the coordinator is faulty
+    instead and has misrepresented the suspected particant's protocol messages.
+    It is also possible that protocol message was simply not transmitted
+    correctly.
+
+    Attributes:
+        participant (int): Index of the suspected participant.
+    """
```
real-or-random, line 41: I went ahead and added a docstring here, because I felt this (and `FaultyCoordinatorError`) deserve one. Then I realized that this won't show in the README then.

So far, we had documented exceptions only in the functions that raise them. Here's what we could do:
1. Keep documenting exceptions only in the functions that raise them. That means we'll need to repeat the essence of the string here in many functions, so that's a bit repetitive.
2. Like 1, but only include it in one function X and then refer to X in docstrings of other docstrings. That's doable but a bit messy.
3. Keep the docstrings here, and make sure they're included in the README. Of course, they'd still be mentioned briefly in the raising functions. 

I think that approach 3 is appropriate now with IA given the number of exceptions and the subtleties in their meanings, but in the end, any approach works. What do you think?

python/chilldkg_ref/simplpedpop.py
real-or-random: I pushed a commit (still WIP, tests not adjusted). Can you have a look to see if this approach is good? The catching and reraising of exceptions is perhaps a bit hard to read, or at least inelegant.

python/chilldkg_ref/simplpedpop.py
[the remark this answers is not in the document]
jonasnick: However, a comment may make this even more confusing. Nevermind :shrug:

python/chilldkg_ref/util.py
jonasnick: I'd choose with approach 3.

python/chilldkg_ref/simplpedpop.py
jonasnick: I don't think the catching and reraising looks too bad. A rust-like `Result` type would imo look cleaner in general, but we decided that exceptions are more idiomatic python.

Maybe we can add a comment that says that we are catching an exception that carries state and raising an exception that carries new state because that seems to be the main thing that may be more difficult to comprehend.

Have you thought about putting more state into the blame state such that we can remove arguments from `participant_blame`? This would simplify the API considerably.

The downside is that participants keep redundant state in their memory, but actually not in all scenarios. If implementers consider the round to be finally failed if `participant_step2` fails and never retry the step even after blaming, they could only keep the blame state and forget all other state.

python/chilldkg_ref/util.py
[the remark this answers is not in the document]
real-or-random: > I may want to improve further this by adding a subclass InvalidHostpubkeyError for the common case when a function gets an invalid hostpubkey.

Marking as unresolved again as a reminder for this small TODO.

python/chilldkg_ref/simplpedpop.py
jonasnick: Is there anything that can go wrong if they retry with new step2 inputs?

real-or-random: > Is there anything that can go wrong if they retry with new step2 inputs?

I don't see a concrete problem. If you're an honest participant, retrying with new inputs *should* be safe. But I tend to think that we should rather err on the side of caution here. Retrying would make the space of possible executions much larger, and then it's easier to overlook a security issue. If one anyway needs manual intervention because a bugfix needs to be deployed or broken network needs to be fixed, then restarting the session should be acceptable.

I think it also depends a bit on the perspective.  The fact that there is a second could mean that some other participant had a bug, and perhaps retrying is not safe for that participant then.

jonasnick: Makes sense. In that case, it's less of a problem that BlameData contains secret data, because it should not be relevant anymore (as you pointed out).

real-or-random: > > Have you thought about putting more state into the blame state such that we can remove arguments from `participant_blame`? This would simplify the API considerably.
> 
> Good point. Ideally, blame only needs the CoordinatorBlameMsg and the BlameState.

Done, but it's still WIP. The simplpedpop tests work, but I need to thread the same logic to the encpedpop and chilldkg tests.  Will let you know once it's ready for review.

jonasnick: Had a look at the new BlameState design and it seems good so far.

python/secp256k1proto/secp256k1.py
```diff
@@ -15,6 +15,9 @@
 * G: the secp256k1 generator point
 """
 
+from typing_extensions import Self
```
jonasnick, line 18: Is there a way to work around this dependency? It's not part of python's standard library.

So e.g., `python ./python/example.py` fails with `ModuleNotFoundError: No module named 'typing_extensions'` using a vanilla python installation (tested with python 3.11 and python 3.12). We don't have to explicitly install this module in CI because `pydoc-markdown` provides it.

real-or-random: Ha, hm, `Self` is in the standard library in 3.13. The idea of the `typing_extension` module is that it provides it for lower versions. But yes, I didn't consider that this is an external dependency and we'd need a package manager. Perhaps we can copy the one def for `Self` from the module, but I have no idea if this is easy. 

But now that I think about it, we'll need a package manager anyway as soon as we package secp256k1-proto separately. That's unfortunate. Still, I tend to think that this is still better than adding the entire secp256k1-proto to the BIPs repo. I don't know, neither way feels great.

python/example.py
```diff
@@ -65,15 +73,41 @@ async def receive(self):
...
     random = random_bytes(32)
     state1, pmsg1 = participant_step1(hostseckey, params, random)
+
+    # The following code simulate a faulty participant that sends a single
+    # incorrect share with probability 1/2. Of course, this should not be part
+    # of a real implementation.
+    if blame:
+        # Flip a coin to decide if we send incorrect shares.
+        faulty = randint(0, 1)
+        print(faulty)
```
jonasnick, line 91: Is this intentional?

jonasnick (approved):
ACK 29adb702c7a616fad41b95c9d3cc469bd50f51bb

python/example.py
jonasnick: fixed in follow-up commit

python/secp256k1proto/secp256k1.py
jonasnick: We discussed offline that requiring an external dependency is fine if secp256k1proto becomes a package. For now, secp256k1proto could also be a submodule instead of a package. Whatever is easier until we have a final deisionon this.

python/chilldkg_ref/util.py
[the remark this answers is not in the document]
real-or-random: > I may want to improve further this by adding a subclass InvalidHostpubkeyError for the common case when a function gets an invalid hostpubkey.

After further consideration, I think this is overkill.

Issue #29, opened 2024-06-28
Should the coordinator check inputs as far as possible?


real-or-random:
This is related to #9, but only somewhat.

At the moment, the coordinator mostly just relays messages. The verification of the messages is left to the participants, who need to perform it anyway. The coordinator could also check messages. Does this provide any advantages?

DarkWindman:
I agree with @jonasnick that it would increase complexity and lead to duplicated checks. 

Replacing the error handling with a model where only the coordinator can be blamed may seem reasonable, but since the coordinator cannot be fully trusted, this approach becomes problematic. In general case, such a check would be redundant. 
However, as mentioned above, implementers can always add this check at a higher level of the implementation.

real-or-random:
Ok, let's not do it.

PR #53, opened 2024-11-27
Rework blame exceptions

real-or-random:
Collecting some early feedback here.

### DIFF
--- a/python/chilldkg_ref/util.py
+++ b/python/chilldkg_ref/util.py
@@ -34,7 +34,7 @@ class FaultyParticipantError(ProtocolError):
 
     This exception is raised only by the coordinator code. Some faulty behavior
     by participants will be detected by the other participants instead.
-    See FaultyParticipantOrCoordinatorError for details.
+    See `FaultyParticipantOrCoordinatorError` for details.
 
     Attributes:
         participant (int): Index of the faulty participant.
@@ -64,7 +64,7 @@ class FaultyParticipantOrCoordinatorError(ProtocolError):
 
     This exception is raised only by the participant code. Some faulty behavior
     by participants will be detected by the coordinator instead. See
-    FaultyParticipantError for details.
+    `FaultyParticipantError` for details.
 
     Attributes:
         participant (int): Index of the suspected participant.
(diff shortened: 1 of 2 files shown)


python/chilldkg_ref/util.py
```diff
@@ -27,14 +27,31 @@ class ProtocolError(Exception):
...
+    either of the following is true:
+
+      - The suspected participant is faulty.
+      - The coordinator is faulty (and has not processed the processed the
+        encrypted secret shares sent by the suspected participant correctly).
+
+    In other words, if this error is raised by an (honest) participant, then
+    this does not necessarily imply that the suspected participant is faulty. It
+    is also possible that the coordinator is faulty instead and has incorrectly
+    framed the suspected participant.
```
real-or-random, line 54: I think this is an improvement, but I don't like the case distinction in the text. In particular, I don't like that this has "Faulty...Coordinator" in the name but could be raised by the coordinator itself. I plan to split this exception into `FaultyParticipantOrCoordinatorError` (raised only by participants in case of invalid shares) and `FaultyParticipantError` (raised only by the coordinator, #29). Concept ACK?

python/chilldkg_ref/util.py
```diff
@@ -49,13 +66,12 @@ class FaultyCoordinatorError(ProtocolError):
...
-    message. This error does not necessarily imply that the coordinator is
-    faulty. It is also possible that a protocol message from the coordinator was
-    simply not transmitted correctly.
+    message. If the protocol message deemed invalid was transmitted correctly,
+    this implies that the coordinator is faulty.
     """
 
 
-class UnknownFaultyPartyError(ProtocolError):
+class UnknownFaultyParticipantOrCoordinatorError(ProtocolError):
```
real-or-random, line 74: I like this explicitness of "Participant  or coordinator" but `UnknownFaultyParticipantOrCoordinatorError` is very long. We could instead replace "Participant or coordinator" by "Party" (also above). Any preference? If you don't have a preference, I'll probably stick to the explicit but long name because "party" could be misunderstood as participant.

python/chilldkg_ref/util.py
jonasnick: Yes. Having the coordinator throw a `FaultyParticipantOrCoordinatorError` is strange.

real-or-random:
Ready for review.

Meanwhile, I'll continue with the docs of the blame functions.

real-or-random:
I accidentally pushed most of this to master already in https://github.com/BlockstreamResearch/bip-frost-dkg/commit/f3a65a07f1969d6fad615fbebe07e2af0883705e. Please review there.

jonasnick (commented):
ACK

PR #57, opened 2024-12-04
Prevent malicious taproot commitment (Approach 1)

### LINKED ISSUE #41: Warning about using the DKG output directly on-chain
Unlike MuSig2, the FROST group public key is not randomized. Therefore, a malicious party could add a script path to their polynomial during the DKG, and still provide a valid proof-of-knowledge and VSS. It is necessary that an unspendable script path be added as [suggested](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-23) by BIP341. It might be worth warning users in this BIP that they should not use the DKG output directly on chain for this reason.

jonasnick:
Fixes #41

See https://github.com/BlockstreamResearch/bip-frost-dkg/issues/41#issuecomment-2517779078 for discussion of the alternatives.

Before this is ready, I'll need to fix the formatting and type errors and add some text to the README.md.


python/chilldkg_ref/simplpedpop.py
```diff
@@ -87,6 +88,16 @@ class DKGOutput(NamedTuple):
     pubshares: List[bytes]
 
 
+# TODO: add comment
+def vss_invalid_taproot_commit(vss: VSSCommitment) -> Tuple[VSSCommitment, Scalar]:
```
jonasnick, line 92: nit: should be defined after assemble_sum_coms

python/chilldkg_ref/vss.py
```diff
@@ -74,6 +74,32 @@ def commitment_to_secret(self) -> GE:
     def commitment_to_nonconst_terms(self) -> List[GE]:
         return self.ges[1 : self.t()]
 
+    def invalid_taproot_commit(self) -> Tuple[VSSCommitment, Scalar]:
+        # Create VSS commitment that Taproot-commits to a message that is
+        # invalid for BIP 341 script path spends.
```
real-or-random, line 79:
```suggestion
        # Return a modified VSS commitment whose corresponding  Taproot (BIP341) output key has an unspendable script path
```

It's hard to write a concise description here. I think this is the best I can come up with

python/chilldkg_ref/vss.py
```diff
@@ -74,6 +74,32 @@ def commitment_to_secret(self) -> GE:
...
+        # Taproot commitment to a script path into the summed VSS commitment during
+        # the DKG protocol. If the resulting threshold public key was used directly
+        # in a BIP 341 Taproot output, the malicious participant would be able to
+        # spend the output using their hidden script path.
+        #
+        # The function returns the updated VSS commitment and the tweak `t` which
+        # must be added to all secret shares of the commitment.
+        pk = self.commitment_to_secret()
+        secshare_tweak = Scalar.from_bytes(
+            tagged_hash_bip_dkg("invalid Taproot commitment", pk.to_bytes_compressed())
```
real-or-random, line 98: Should we use the exact method suggested by BIP341, simply to avoid creating yet another way of doing this? https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_ref-23-0

real-or-random (commented):
concept ACK, this is a clean approach

python/chilldkg_ref/vss.py
```diff
@@ -74,6 +74,32 @@ def commitment_to_secret(self) -> GE:
...
 
+    def invalid_taproot_commit(self) -> Tuple[VSSCommitment, Scalar]:
+        # Create VSS commitment that Taproot-commits to a message that is
+        # invalid for BIP 341 script path spends.
+        #
+        # Specifically, for a VSS commitment `com`, we have:
+        # `com.add_invalid_taproot_commitment().commitment_to_secret() = com.commitment_to_secret() + t*G`.
+        #
+        # The tweak `t` commits to a message chosen such that the new
+        # `commitment_to_secret()` is unspendable via a BIP 341 Taproot script
```
real-or-random, line 85: consistency nit: (multiple times)
```suggestion
        # `commitment_to_secret()` is unspendable via a BIP341 Taproot script
```

jonasnick: Should we instead add a space between "BIP" and the number everywhere? This is how the old and new BIP 2 refers to BIPs.

python/chilldkg_ref/vss.py
jonasnick: What exactly do you think is the problem with the current description? Happy for suggestions, but "corresponding Taproot output" sounds vague compared to the current wording. In particular, it does not make clear that the function creates a Taproot commitment.

python/chilldkg_ref/simplpedpop.py
```diff
@@ -286,8 +291,9 @@ def coordinator_step(
     cmsg = CoordinatorMsg(coms_to_secrets, sum_coms_to_nonconst_terms, pops)
 
     sum_coms = assemble_sum_coms(coms_to_secrets, sum_coms_to_nonconst_terms)
-    threshold_pubkey = sum_coms.commitment_to_secret()
-    pubshares = [sum_coms.pubshare(i) for i in range(n)]
+    (sum_coms_tweaked, secshare_tweak) = sum_coms.invalid_taproot_commit()
```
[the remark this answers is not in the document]
jonasnick: fixed

python/chilldkg_ref/vss.py
jonasnick: Ok, yeah, done.

python/chilldkg_ref/vss.py
real-or-random: Sure, happy with either way.

jonasnick:
Added text to the README. "Ready for review" now.

python/chilldkg_ref/vss.py
jonasnick: All fair points. I am using a variant of your last suggestion now. Wdyt?

python/chilldkg_ref/vss.py
```diff
@@ -74,6 +75,31 @@ def commitment_to_secret(self) -> GE:
...
         return self.ges[1 : self.t()]
 
+    def invalid_taproot_commit(self) -> Tuple[VSSCommitment, Scalar]:
+        # Return a modified VSS commitment such that the threshold public key
+        # generated from it has an unspendable BIP 341 Taproot script path.
+        #
+        # Specifically, for a VSS commitment `com`, we have:
+        # `com.add_invalid_taproot_commitment().commitment_to_secret() = com.commitment_to_secret() + t*G`.
+        #
+        # The tweak `t` commits to an empty message which is invalid for BIP 341
```
real-or-random, line 85: This should probably say that we use the method suggested in the BIP.

python/chilldkg_ref/vss.py
```diff
@@ -74,6 +75,31 @@ def commitment_to_secret(self) -> GE:
...
+        # Taproot script spends.
+        #
+        # This prevents a malicious participant from secretly inserting a _valid_
```
real-or-random, line 88: nit: 
```suggestion
        # This prevents a malicious participant from secretly inserting a *valid*
```

python/chilldkg_ref/vss.py
```diff
@@ -74,6 +75,31 @@ def commitment_to_secret(self) -> GE:
...
+        # Taproot commitment to a script path into the summed VSS commitment during
+        # the DKG protocol. If the resulting threshold public key was used directly
+        # in a BIP 341 Taproot output, the malicious participant would be able to
+        # spend the output using their hidden script path.
+        #
+        # The function returns the updated VSS commitment and the tweak `t` which
+        # must be added to all secret shares of the commitment.
```
real-or-random, line 95: "secret shares committed to in the commitment."? feel free to ignore.

real-or-random (commented):
ACK mod nits

python/chilldkg_ref/vss.py
real-or-random: looks great

python/chilldkg_ref/vss.py
jonasnick: Ignored ;) the VSS commitment naturally "has" secret shares, namely those that pass `verify_secshare`. So I think the existing version is clear enough and less complex.

python/chilldkg_ref/vss.py
jonasnick: Good point. done

python/chilldkg_ref/vss.py
jonasnick: done

python/chilldkg_ref/vss.py
real-or-random: good point

PR #59, opened 2024-12-04
Prevent malicious taproot commitment (Approach 3)

### LINKED ISSUE #41: Warning about using the DKG output directly on-chain

jonasnick:
Fixes #41

See https://github.com/BlockstreamResearch/bip-frost-dkg/issues/41#issuecomment-2517779078 for discussion of the alternatives.

Before this is ready, I'll need to fix the formatting and type errors and add some text to the README.md.


jonasnick:
Alternative PR was merged.

PR #60, opened 2024-12-05
text: visually stress usage requirements ("should"/"must")

### LINKED ISSUE #56: Highlight usage recommendations
We **should** highlight things that implementors should do, e.g., by making the word "should" bold.

jonasnick:
Fixes #56.

This should also make it easier to grep important usage guidelines by grepping for bold (or italic, see below) text.

Asterisks in bold text pydoc strings are stripped such that they only appear italic. This is explicitly implemented in `update-pydoc.sh`. Is there a particular reason for this @real-or-random ?

I also removed the capitalization from two occurences of "MUST NOT" in a pydoc string for consistency. Because these strings are now only italic, they're actually less visually highlighted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,9 +450,9 @@ should also read [Section "Internals of ChillDKG"](#internals-of-chilldkg).
 ChillDKG is designed for usage with the FROST Schnorr signature scheme,
 and its security depends on the specifics of FROST.
 We stress that ChillDKG is not a general-purpose DKG protocol,[^no-simulatable-dkg]
-and combining it with other threshold cryptographic schemes,
-e.g., threshold signature schemes other than FROST, or threshold decryption schemes
-requires careful further consideration, which is not endorsed or in the scope of this document.
+and **must not** be combined with other threshold cryptographic schemes,
+e.g., threshold signature schemes other than FROST, or threshold decryption schemes,
+without careful further consideration, which is not in the scope of this document.
 
 [^no-simulatable-dkg]: As a variant of Pedersen DKG, ChillDKG does not provide simulation-based security [GJKR07](https://doi.org/10.1007/s00145-006-0347-3). Roughly speaking, if ChillDKG is combined with some threshold cryptographic scheme, the security of the combination is not automatically implied by the security of the two components. Instead, the security of every combination must be analyzed separately. The security of the specific combination of SimplPedPop (as the core building block of ChillDKG) and FROST has been analyzed [CGRS23](https://eprint.iacr.org/2023/899).
 
@@ -490,7 +490,7 @@ they can request it from any other participants or the coordinator.
 Moreover, the recovery data contains secrets only in encrypted form and is self-authenticating
 so that it can, in principle, be stored with an untrusted third-party backup provider.
 
-Users should be aware that the session parameters (the threshold and the host public keys) and public parts of the DKG output (the threshold public key and the public shares) can be inferred from the recovery data, which may constitute a privacy issue.
+Users **should** be aware that the session parameters (the threshold and the host public keys) and public parts of the DKG output (the threshold public key and the public shares) can be inferred from the recovery data, which may constitute a privacy issue.
 To eliminate this issue, users can encrypt the recovery data using an encryption key derived from their host secret key before publishing the data.
 Recovery from encrypted data requires only the participant's host secret key, with no additional secrets needed.
 This BIP does not specify the encryption scheme.
@@ -516,7 +516,7 @@ The recovery data can, e.g., be attached to the first request to initiate a FROS
 
 An important implication of the above is that anyone who uses the threshold public key,
 and thereby relies on the participants' ability to participate in signing sessions,
-is responsible for ensuring that the participants have already deemed the DKG session successful,
+**must** ensure that the participants have already deemed the DKG session successful,
 or at least, that the recovery data will be available to convince any stuck participants of the success of the DKG session.
 
 For an example of what could go wrong,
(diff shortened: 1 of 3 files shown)


real-or-random (commented):
Concept ACK

real-or-random:
> Asterisks in bold text pydoc strings are stripped such that they only appear italic. This is explicitly implemented in `update-pydoc.sh`. Is there a particular reason for this @real-or-random ?

IIRC This is because the automatically generated heading: "Arguments:", "Returns:", "Raises:", and "Attributes:" will be bold otherwise, and then they're hard to distinguish from the higher-order headings, at least in GitHub's rendering. I suggest we change the script to replace `**` with `*` only for these specific strings. I agree that bold is best for stressing "must"/"should".

README.md
```diff
@@ -442,7 +442,7 @@ and its security depends on the specifics of FROST.
 We stress that ChillDKG is not a general-purpose DKG protocol,[^no-simulatable-dkg]
 and combining it with other threshold cryptographic schemes,
 e.g., threshold signature schemes other than FROST, or threshold decryption schemes
-requires careful further consideration, which is not endorsed or in the scope of this document.
+**requires** careful further consideration, which is not endorsed or in the scope of this document.
```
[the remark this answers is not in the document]
jonasnick: Agree, but I would highlight "must not" instead of just "must".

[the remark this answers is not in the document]
jonasnick: done

jonasnick:
I see. I excluded must, should, must not and should not using python + regexes. According to LLMs `sed` cannot really do that.

README.md
```diff
@@ -440,9 +440,9 @@ should also read [Section "Internals of ChillDKG"](#internals-of-chilldkg).
 ChillDKG is designed for usage with the FROST Schnorr signature scheme,
 and its security depends on the specifics of FROST.
 We stress that ChillDKG is not a general-purpose DKG protocol,[^no-simulatable-dkg]
-and combining it with other threshold cryptographic schemes,
-e.g., threshold signature schemes other than FROST, or threshold decryption schemes
-requires careful further consideration, which is not endorsed or in the scope of this document.
+and **must not** be combined with other threshold cryptographic schemes,
```
real-or-random, line 443: do you think "should not" would be more appropriate here?

README.md
```diff
@@ -641,9 +641,9 @@ A `SessionParams` tuple holds the common parameters of a DKG session.
 - `hostpubkeys` - Ordered list of the host public keys of all participants.
 - `t` - The participation threshold `t`.
   This is the number of participants that will be required to sign.
-  It must hold that `1 <= t <= len(hostpubkeys)` and `t <= 2^32 - 1`.
+  It **must** hold that `1 <= t <= len(hostpubkeys)` and `t <= 2^32 - 1`.
```
real-or-random, line 644: Do we need to stress this one? I think this boils down to the question whether we may write "must" without making it bold. I think it will be appropriate here. (The inequalities must hold, but nothing bad happens if they don't. You get an exception and that's it.)

README.md
```diff
@@ -656,8 +656,8 @@ A `SessionParams` tuple holds the common parameters of a DKG session.
   where the participants need to obtain authentic individual public
   keys of the other participants to generate an aggregated public key.
 
-  All participants and the coordinator in a session must be given an identical
-  `SessionParams` tuple. In particular, the host public keys must be in the
+  All participants and the coordinator in a session **must** be given an identical
+  `SessionParams` tuple. In particular, the host public keys **must** be in the
```
real-or-random, line 660: Same here, I don't think these need to be bold.

README.md
```diff
@@ -506,21 +506,21 @@ The recovery data can, e.g., be attached to the first request to initiate a FROS
...
 and thus (besides the untrusted coordinator) the only one who knows the recovery data.
-If the recovery data is lost now because this participant's permanent storage crashes,
+If the recovery data is lost now because this participant's permanent storage fails,
 the other participants cannot be convinced to deem the DKG session successful
 (without the help of the untrusted coordinator)
 and so the funds will be lost.
 
 Thus, anyone who intends to use the threshold public key
-should first obtain explicit confirmations from all participants that they have deemed the DKG session successful,
+**should** first obtain explicit confirmations from all participants that they have deemed the DKG session successful,
```
real-or-random, line 523: Do you think we should make this a "must" in terms of the badness of the consequences and given that say "must" in line 509?

README.md
```diff
@@ -531,7 +531,7 @@ Alternatively, the user could check all `n` devices when generating a receiving
 which constitutes the first use of the threshold public key.
 
 If a recovering party (see [Backup and Recovery](#backup-and-recovery)) cannot (re-)obtain confirmations,
-this simply means they should stop using the threshold public key going forward,
+this simply means they **should** stop using the threshold public key going forward,
```
real-or-random, line 534: Same here, make this a "must"?

README.md
jonasnick: Hm no. I don't think there exists circumstances when using the BIP in other systems without careful consideration is acceptable.

README.md
jonasnick: If the session params is not identical, then there's no authenticity of the host public keys, so I don't see how it's related to your previous comment. Maybe the second one doesn't need to be stressed.

I'm thinking we should change "be given" to "have" because it's clearer.

README.md
jonasnick: As per line 509, confirmation messages are unnecessary if it can be ensured that the recovery data is available.

README.md
jonasnick: Same here, recovery data may be available.

README.md
jonasnick: Ok I guess you're suggesting to distinguish by boldness.

README.md
real-or-random: Indeed.

README.md
real-or-random: yes

README.md
real-or-random: > If the session params is not identical, then there's no authenticity of the host public keys, so I don't see how it's related to your previous comment. Maybe the second one doesn't need to be stressed. 

Hm, I think this is more involved and we're both a bit correct and incorrect. These are ultimately instructions for the caller, and the caller can only ensure that the local call is correct, and has no influence on what other parties do. The requirement to check public keys is also stated a few lines above. 

I think the reason why I wrote this sentence is simply to prepare the next sentences that talk about the order. And what I had in mind here this is that the session will simply fail if different honest parties have different inputs. 
 
I suggest not saying too much about the SessionParams being equal here. Anything we could say be confusing and distract from the requirement to check keys (e.g., if we say "the protocol will fail if parties use different SessionParams", someone could read this as "You don't need to check the pubkeys, the protocol will do it for you."). And I think it's implicitly expected that the protocol doesn't work if parties have different inputs (including t), so I don't think we'll need to say this.

So I suggest we mention only the requirement to have the same order. 

Here's a possible way, feel free to adapt:
"A DKG session will fail if the participants and the coordinator in a session don't have the `hostpubkeys` in the same order. This will make sure that honest participants agree on the order as part of the session, which is useful if the order carries an implicit meaning in the application (e.g., if the first `t` participants are the primary participants for signing and the others are fallback participants)."

This one also avoids the "must" entirely. What do you think?

README.md
jonasnick: done

README.md
jonasnick: Ok I see. Picked your suggestion.

PR #62, opened 2024-12-09
python: Replace SessionNotFinalizedError by "Faulty" Errors

real-or-random:


README.md
```diff
@@ -890,6 +892,24 @@ def coordinator_finalize(state: CoordinatorState, pmsgs2: List[ParticipantMsg2])
 
 Perform the coordinator's final step of a ChillDKG session.
 
+If this function returns properly (without an exception), then the
+coordinator deems the DKG session successful. The returned `CoordinatorMsg2`
+is supposed to be sent to all participants, who are supposed to pass it
+as input to the `participant_finalize` function. It is, however, possible
+that some participants participants pass a wrong and invalid message to
```
jonasnick, line 899:
```suggestion
that some participants pass a wrong and invalid message to
```

jonasnick (commented):
In short, what is the rationale for this change?

real-or-random:
> In short, what is the rationale for this change?

IA requires us to raise a `Faulty*` exception for every failure case (so that we attribute blame to someone in every failure case), but currently we raise `SessionNotFinalizedError` if the received certificate fails to verify. [1]

This PR replaces `SessionNotFinalizedError` by the corresponding exception, namely either `FaultyParticipantOrCoordinatorException` (if verification fails in the participant) or `FaultyParticipantException` (if verification fails in the coordinator). This requires some additional logic to distinguish between the two candidate exceptions.

Additionally, I sneaked in some docstring improvements for `coordinator_finalize`, inspired by `participant_finalize`.

README.md
real-or-random: fixed

jonasnick (approved):
ACK

PR #64, opened 2024-12-10
text: mention that index is required for signing

### LINKED ISSUE #38: Index assignment
From the BIP

>All participants and the coordinator in a session must be given an identical SessionParams tuple. In particular, the host public keys must be in the same order. This will make sure that honest participants agree on the order as part of the session, which is useful if the order carries an implicit meaning in the application (e.g., if the first t participants are the primary participants for signing and the others are fallback participants). If there is no canonical order of the participants in the application, the caller can sort the list of host public keys with the KeySort algorithm specified in BIP327 to abstract away from the order.

It's not explicit in the BIP whether the indices are assigned from this ordering. We've discussed this here (https://github.com/siv2r/bip-frost-signing/issues/5#issuecomment-2178705603) so I believe it is, but it might be worth making this explicit, because the index assignment is going to affect whatever signing protocol is compatible with ChillDKG. After the DKG, the indices will either need to be extracted from SessionParams and persisted for use for signing protocols, or the ordering of the keys in SessionParams will need to be persisted so the indices can be derived from them.

In addition, I think I prefer indices derived from long-term identity keys rather than from ordering because I believe that will make it easier to work with dynamic secret sharing protocols in which participants can be added and removed. If indices are derived from ordering, then if a participant is added or removed, the ordering can no longer be used, and instead the indices need to be maintained as a map to the long-term identity keys (or some other identity for each participant). Whereas with derivation from identity keys, the index distribution is stateless and doesn't depend on an ordering or a map.

jonasnick:
Fixes #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ We make the following modifications as compared to the original SimplPedPop prop
 
 Our variant of the SimplPedPop protocol then works as follows:
 
-1.  Every participant `i` creates a `t`-of-`n` sharing of a random secret scalar using Feldman Verifiable Secret Sharing (VSS), a variant of Shamir Secret Sharing.
+1.  Every participant `i` (where `i` is an integer `0 <= i < n`) creates a `t`-of-`n` sharing of a random secret scalar using Feldman Verifiable Secret Sharing (VSS), a variant of Shamir Secret Sharing.
     This involves generating random coefficients `a_i[0], ..., a_i[t-1]` of a polynomial `f_i` of degree `t-1` in the scalar group:
 
     ```
@@ -472,6 +472,9 @@ If a session ChillDKG returns an output to a participant or the coordinator,
 then we say that this party *deems the protocol session successful*.
 In that case, the DKG output is a triple consisting of a *secret share* for participating in FROST signing sessions (individual to each participant, not returned to the coordinator), the *threshold public key* representing the `t`-of-`n` policy of the group (common to all participants and the coordinator), and a list of `n` *public shares* for verification of individual contributions to a FROST signing session (common to all participants and the coordinator).
 Moreover, all parties obtain *recovery data* (common to all participants and the coordinator), whose purpose is detailed in the next subsection.
+
+To participate in the FROST signing protocol, signers need their DKG output and their index in the host public key list, although the full list of host public keys is not required for signing.
+Additionally, the set of indices of all participating signers within the host public key list is required to initiate a signing session.
 
 ### Backup and Recovery
 


README.md
```diff
@@ -473,6 +473,9 @@ then we say that this party *deems the protocol session successful*.
 In that case, the DKG output is a triple consisting of a *secret share* for participating in FROST signing sessions (individual to each participant, not returned to the coordinator), the *threshold public key* representing the `t`-of-`n` policy of the group (common to all participants and the coordinator), and a list of `n` *public shares* for verification of individual contributions to a FROST signing session (common to all participants and the coordinator).
 Moreover, all parties obtain *recovery data* (common to all participants and the coordinator), whose purpose is detailed in the next subsection.
 
+To participate in the FROST signing protocol, signers need their DKG output and their index in the host public key list, although the full list of host public keys is not required for signing.
+Additionally, to initiate a signing session, the signer or coordinator must know the indices of all participating signers within the host public key list.
```
real-or-random, line 477: When I read this sentence, I asked myself "which signer (or coordinator)"? I think who exactly will need it depends a bit on i) which FROST variant is used (Lategrange or not), and ii) how the application organizes things. I assume this is what you want to convey with "signer *or* coordinator"?

Perhaps that's a bit simpler then:
```suggestion
Additionally, the set of indices of all participating signers within the host public key list is required to initiate a signing session.
```

If you stick to the current text, I think s/signer/participant is a bit more consistent. While a participant will also be a signer, we currently stick to "participant" throughout the text.

real-or-random (commented):
So you're suggesting not to change the DKGOutput to include the index, right? Can you elaborate on why not?

jonasnick:
> So you're suggesting not to change the DKGOutput to include the index, right? Can you elaborate on why not?

Correct. The reason to *include* the index in the DKG output would be that this would make the DKG output contain sufficient information for signing. However, if the signer wants to initiate a signing session, this DKG output would not be sufficient because the signer needs to know also the indices of the participating signers. Hence, even for that DKG output the signer (and coordinator) would have to have a way to get the index of a particular hostpubkey from the list (if they want to initiate a signing session).

Also, finding the index of a hostpubkey in the list is trivial (and it's in some sense already input of the DKG).

README.md
jonasnick: Agreed. Replaced the sentence with your suggestion.

jesseposner:
>Also, finding the index of a hostpubkey in the list is trivial (and it's in some sense already input of the DKG).

Is this because the ordering of the hostpubkeys in the list determines the index assignment?

jonasnick:
Yes

real-or-random:
> Also, finding the index of a hostpubkey in the list is trivial (and it's in some sense already input of the DKG).

Okay, that makes sense. Also with the signing BIP in mind, which uses indices instead of arbitrary identifers.

If the signing BIP were to use arbitrary identifiers instead, we'd probably want to output our choice of identifiers. But if everything uses indices, then it's okay to leave this implicit. (The only caveat is that signing takes as input 32-byte arrays that will be interpreted as integer indices. I think it's a bit more natural to use integers directly, but perhaps this could be changed in the signing BIP)

real-or-random:
> (The only caveat is that signing takes as input 32-byte arrays that will be interpreted as integer indices. I think it's a bit more natural to use integers directly, but perhaps this could be changed in the signing BIP)

Just noticed that this is tracked in https://github.com/siv2r/bip-frost-signing/issues/9 already.

PR #70, opened 2024-12-16
Add unfinalized session warning to participant_step2

### LINKED ISSUE #63: Reorganize warning about unfinalized sessions
The warning that other participants could be finalized even if we're not is currently in `participant_finalize`.

But it should also be in `participant_step2` for the case that the participant aborts during step2 and finalize, e.g., due to a network failure.

jonasnick:
Fixes #63 

Copied parts of the wording from the `participant_finalize` docs. It's a bit strange that the text refers to the reader/user both as "you" and "us". Not sure if that was intentional,

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -490,6 +490,18 @@ def participant_step2(
 ) -> Tuple[ParticipantState2, ParticipantMsg2]:
     """Perform a participant's second step of a ChillDKG session.
 
+    **Warning:**
+    After sending the returned message to the coordinator, this participant
+    **must not** erase the hostseckey, even if this participant does not receive
+    the coordinator reply needed for the `participant_finalize` call. The
+    underlying reason is that some other participant may receive the coordinator
+    reply, deem the DKG session successful and use the resulting threshold
+    public key (e.g., by sending funds to it). If the coordinator reply remains
+    missing, that other participant can, at any point in the future, convince
+    this participant of the success of the DKG session by presenting recovery
+    data, from which this participant can recover the DKG output using the
+    `recover` function.
+
     Arguments:
         hostseckey: Participant's long-term host secret key (32 bytes).
         state1: The participant's session state as output by
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -557,12 +557,13 @@ def participant_finalize(
 
     **Warning:**
     Changing perspectives, this implies that, even when obtaining an exception,
-    you **must not** conclude that the DKG session has failed, and as a
-    consequence, you **must not** erase the hostseckey. The underlying reason is
-    that some other participant may deem the DKG session successful and use the
-    resulting threshold public key (e.g., by sending funds to it). That other
-    participant can, at any point in the future, wish to convince us of the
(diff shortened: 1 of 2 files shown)


python/chilldkg_ref/chilldkg.py
```diff
@@ -490,6 +490,16 @@ def participant_step2(
...
 
+    **Warning:**
+    After sending the returned message to the coordinator, you **must not**
+    erase the hostseckey, even if you don't receive the coordinator reply needed
+    for the `participant_finalize` call. The underlying reason is that some
+    other participant may receive the coordinator reply, deem the DKG session
+    successful and use the resulting threshold public key (e.g., by sending
+    funds to it). If the coordinator reply remains missing, that other
+    participant can, at any point in the future, convince us of the success of
+    the DKG session by presenting recovery data to us.
```
real-or-random, line 501: 1. suggestion:
"(...) recovery data to us, from which we can recover the DKG outputs using the `recover` function."

    In `participant_finalize`, this is how the term "recovery data" is briefly explained in the paragraph above the warning. 


2. You changed "wish to convince us" to "convince us". I agree that's better because it's simpler. I suggest changing it in `participant_finalize` too.

real-or-random (commented):
ACK

real-or-random:
> It's a bit strange that the text refers to the reader/user both as "you" and "us". Not sure if that was intentional,

Yeah, this is the only place where we speak to the caller directly, so it's a confusing style. I had noticed this, too, and I don't like it, but I was too lazy to fix it so far.

There are two obvious ways: 
 - Use passive voice, and replace "us" by "this participant"
 - Replace "you" by "this participant" (or "the participant" if appropriate)

The second is good if you ask me. Here's how it would look like (modulo my nits above and line lengths) :

    After sending the returned message to the coordinator, this participant **must not**
    erase the hostseckey, even if the participant doesn't receive the coordinator reply needed
    for the `participant_finalize` call. The underlying reason is that some
    other participant may receive the coordinator reply, deem the DKG session
    successful and use the resulting threshold public key (e.g., by sending
    funds to it). If the coordinator reply remains missing, that other
    participant can, at any point in the future, convince this participant of the success of
    the DKG session by presenting recovery data to us."

python/chilldkg_ref/chilldkg.py
jonasnick: Done

jonasnick:
> Here's how it would look like

Switched to that version except that I replaced the remaining us (last word) to "this participant".

python/chilldkg_ref/chilldkg.py
```diff
@@ -490,6 +490,18 @@ def participant_step2(
 ) -> Tuple[ParticipantState2, ParticipantMsg2]:
     """Perform a participant's second step of a ChillDKG session.
 
+    **Warning:**
+    After sending the returned message to the coordinator, this participant
+    **must not** erase the hostseckey, even if this participant don't receive
```
real-or-random, line 495:
```suggestion
    **must not** erase the hostseckey, even if this participant does not receive
```

real-or-random (commented):
ACK

python/chilldkg_ref/chilldkg.py
jonasnick: Sorry, fixed

PR #71, opened 2024-12-16
text: Explain usage aspects of blaming

real-or-random:


README.md
```diff
@@ -135,6 +135,7 @@ In summary, we aim for the following design goals:
  - **Broad applicability**:  ChillDKG supports a wide range of scenarios, from those where the signing devices are owned and connected by a single individual to those where multiple owners manage the devices from distinct locations.
  - **Simple backups**: ChillDKG allows recovering the DKG output using the host secret key and common recovery data shared among all participants and the coordinator. This eliminates the need for session-specific backups, simplifying user experience.
  - **Untrusted coordinator**: Like FROST, ChillDKG uses a coordinator that relays messages between the participants. This simplifies the network topology, and the coordinator additionally reduces communication overhead by aggregating some of the messages. A malicious coordinator can force the DKG to fail but cannot negatively affect the security of the DKG.
+ - **Blaming functionality**: If a ChillDKG session aborts, it is possible to identify and blame a single party responsible for the failure (assuming the network, and, depending on the circumstances, the coordinator, is reliable).
```
jonasnick, line 138: I think we have a paragraph for each of the features in the section above and this list is just a summary of the above section (we also say "in summary ..."). But it's also fine to leave as is.

README.md
```diff
@@ -144,14 +145,13 @@ and due to its low overhead, we recommend ChillDKG even if secure communication
 
 #### Why Robustness is Not a Goal
 
-As a consequence of its design goals, ChillDKG does not provide robustness, i.e., the protocol is not guaranteed to succeed in the presence of malicious or faulty participants.
+If a ChillDKG session fails due to the participants or the coordinator deviating from the protocol,
+any aborting party will be able to identify and blame a single party responsible for the failure (assuming the network, and, depending on the circumstances, the coordinator, is reliable).
+
+However, ChillDKG does not provide full robustness, i.e., the protocol is not designed to succeed in the presence of faulty participants.
```
jonasnick, line 151: Why not just say robustness instead of full robustness? If we stick to "full robustness", what version of robustness do we have in ChillDKG? Also, we should change "Adding robustness to ChillDKG" to "Adding full robustness to ChillDKG".

README.md
```diff
@@ -553,18 +556,34 @@ e.g., stop sending additional funds to addresses derived from it.
 it will still be possible to spend the funds,
 and even recovered participants can participate in signing sessions.)
 
+### Blaming Functionality
```
jonasnick, line 559: Nice. The section is precise & understandable.

README.md
```diff
@@ -467,6 +467,9 @@ Each participant has a point-to-point communication link to the *coordinator*
 
 If there is no dedicated coordinator, one of the participants can act as the coordinator.
 
+Each participant and the coordinator, and the communication links may either be *honest*, i.e., reliable and adhering to the protocol, or *faulty*, i.e., controlled by an attacker or unreliable (e.g., due to software bugs).
```
jonasnick, line 470: Is the plan to replace "malicious" with "faulty" everywhere?

real-or-random: I haven't thought this through entirely, but I'd say almost everywhere. (One place where malicious is better IMO is the example where we talk about a "malicious" coordinator. Malicious is more tangible for an example.)

README.md
real-or-random: agreed, will change

README.md
real-or-random: Oh, indeed. I'll try to reorganize this.

real-or-random (commented):
Thanks

real-or-random:
Pushed again for another round of early review.

README.md
```diff
@@ -591,7 +619,21 @@ The participant invokes participant_finalize, which ends the second phase.
...
 Multiple sessions may be run concurrently.
-Whenever a function call fails, the corresponding party will not continue the session.
+
+Whenever a routine fails due to an error, the corresponding party will abort the session and,
+in most cases, blame a participant or the coordinator for the failure of the session.
+However, if a participant aborts during the `participant_step2` routine,
+there may be insufficient information determine another participant to blame.
+In this case, an optional *blame procedure* is available:
+The aborting participant can ask the coordinator for an auxiliary *blame message* (generated via the `coordinator_blame` routine),
+which will allow the participant to blame a specific other participant (via the `participant_blame` routine).
```
real-or-random, line 629: It becomes clearer and clearer to me that the terminology is confusing here. In most cases, a party will be "blamed" using the "blaming functionality" even though the "blame procedure" is not even run. 

I'm considering renaming the "blame procedure" to something like "examination procedure", "investigation procedure", or "debugging procedure". (The "blame message" and the python functions should be renamed accordingly.) What do you think? If you like the idea, any suggestions for the term?

real-or-random:
I'm considering leaving the figure as-is for now. It will probably make sense to add a second figure showing the "blame flow" but this is a stretch goal, I think.

README.md
jonasnick: I agree that this can be confusing. Out of your suggestions I like "investigation" most. The "procedure" feels redundant in many contexts. How about "blame investigation" or "blame resolution"? The latter sounds nice but is probably too strong as there may not be (clear) resolution.

README.md
```diff
@@ -576,11 +604,11 @@ If a participant deems a protocol session successful (see above), then this part
...
 The following figure shows an example execution of the participants and the coordinator.
 Arrows indicate network messages between the participants.
-For simplicity, only one participant is depicted;
-all participants run the identical code and send messages in the same steps.
+For simplicity, only one participant is depicted.
+Unless a participants aborts due to an error, all participants run the same code routines and send messages in the same steps.
 
 ![The diagram shows the message flow between a participant and a coordinator.
-The first of two phases named "Generation of host public keys" involves the participant invoking the function hostpubkey_gen with parameter hostseckey and sending the returned hostpubkey to the coordinator.
+The first of two phases named "Generation of host public keys" involves the participant invoking the routine hostpubkey_gen with parameter hostseckey and sending the returned hostpubkey to the coordinator.
```
jonasnick, line 611: Are you planning to replace "function" with routine in the entire text or only in a particular context? There's still a mention of `"step" functions` below and the API section makes frequent use of "function".

real-or-random: Hm, I like "routine" more in that section because a function can be a mathematical thing, but "function" is the best word for the API docs (this is Python after all and these things are called functions on Python), and consistency matters. I'll revert some, if not all, of these replacements.

README.md
real-or-random: Let's use "investigation" then.

> The "procedure" feels redundant in many contexts. How about "blame investigation" (...)?

It depends on the context. "In this case, an optional *blame investigation* is available:" doesn't really work, but "investigation procedure" will capture exactly what we want to say. I think.

Let me try to work with these ideas a bit, and we can see if we like the result.

jonasnick (approved):
ACK

PR #78, opened 2025-01-20
simplpedpop: remove unused `aux_rand` parameter from `pop_prove`

theStack:
The `aux_rand` parameter in `pop_prove` is currently not used, neither within the function nor at the call-site in `participant_step1` below, so it seems reasonable to remove it. Alternatively, if the intention was really to keep it, it should be passed to `schnorr_sign` and the default value could be set to `random_bytes(32)` instead of all-zeros.

### DIFF
--- a/python/chilldkg_ref/simplpedpop.py
+++ b/python/chilldkg_ref/simplpedpop.py
@@ -35,7 +35,7 @@ def pop_msg(idx: int) -> bytes:
     return idx.to_bytes(4, byteorder="big")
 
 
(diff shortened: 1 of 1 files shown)


real-or-random:
Thanks! I think it's a good idea to use the default `aux_rand`. I think this may be left over from an earlier scheme that was trying to be fully deterministic. In either case, I don't see a reason why the pops shouldn't be randomized.

jonasnick (approved):
ACK

aux_rand was set to random_bytes(32) in https://github.com/BlockstreamResearch/bip-frost-dkg/commit/4e4f6014e928ffc675f276ab4d25592e165aaa5e

PR #79, opened 2025-02-02
secp256k1proto: split `APrimeFE.from_bytes` into checked and wrapping variants

### LINKED ISSUE #77: secp256k1proto:  Split `APrime.from_int` into a wrapping and a checked variant
I think the proper way is to change secp256k1proto to split `APrimeFE.from_int` into a `APrimeFE.from_int_wrapping`, which wraps around, and a `APrimeFE.from_int_checked`, which raises `OverflowError`. (`APrimeFE` is the abstract superclass of `Scalar`).

Then we can use the appropriate function everywhere. (Not sure if we want to touch bip340.py though, I tried to keep this as close as possible to the reference code in the BIP, so it's okay to work with integers there and keep the explicit `%`.)

> > Scalar(int_from_bytes(...)) is also used for generating pads (self_pad and ecdh). I think this is acceptable since we’re generating random numbers,
> 
> Acceptable, but probably better to do this in a consistent way unless you really want to wrap around.

The wrapping there was intentional: Wrapping is not only acceptable when the input is uniformly random (or a hash), it yields simpler code because it removes an error path. This was the rationale for why we wrap in BIP340, after a long discussion.

But the ECDH implementation is supposed to replicate whatever libsecp256k1 does, and [it errors out in case of overflow](https://github.com/bitcoin-core/secp256k1/blob/f79f46c70386c693ff4e7aef0b9e7923ba284e56/src/modules/ecdh/main_impl.h#L71)... 

And then, if we wrap in ECDH, we may as well wrap in `self_pad` for consistency. Sigh.

_Originally posted by @real-or-random in https://github.com/BlockstreamResearch/bip-frost-dkg/issues/76#issuecomment-2592357932_

theStack:
Note that the issue title refers to a `from_int` method, but since the check-or-wraparound discussion mostly concerns the initialization of scalars/fes from byte-strings (e.g. pseudo-random data from tagged hash results), the split-up is applied on the `.from_bytes` method. As suggested in #77, the BIP-340 implementation is not adapted and still applies the wrapping manually (`% GE.ORDER`), in order to keep it close to the reference implementation.

Resolves #77.

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -392,7 +392,7 @@ def deserialize_recovery_data(
     if len(rest) < 32 * n:
         raise ValueError
     enc_secshares, rest = (
-        [Scalar.from_bytes(rest[i : i + 32]) for i in range(0, 32 * n, 32)],
+        [Scalar.from_bytes_checked(rest[i : i + 32]) for i in range(0, 32 * n, 32)],
         rest[32 * n :],
     )
 
--- a/python/chilldkg_ref/encpedpop.py
+++ b/python/chilldkg_ref/encpedpop.py
@@ -3,7 +3,6 @@ from typing import Tuple, List, NamedTuple, NoReturn
 from secp256k1proto.secp256k1 import Scalar, GE
 from secp256k1proto.ecdh import ecdh_libsecp256k1
 from secp256k1proto.keys import pubkey_gen_plain
-from secp256k1proto.util import int_from_bytes
 
 from . import simplpedpop
 from .util import (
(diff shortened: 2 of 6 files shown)


real-or-random:
Concept ACK 

Given my comment in #80, perhaps we should do this:
 - add `from_int_checked` and `from_int_wrapping`
 -  `from_byte_checked` and `from_byte_wrapping` can delegate to the `from_int_*`  methods
 - The constructor can be an alias for `from_int_checked`

real-or-random:
Good point. So we shouldn't touch the constructor. In fact, we use it often within the class, so everything will break if it checks for overflow.

Perhaps we can document in the docstring of the class, or in the one of `__init__`, that it's is low-level and callers should usually prefer one of the factory methods.

real-or-random:
This has been integrated in secp256k1lab

PR #80, opened 2025-02-02
secp256k1proto: deserialize x-only-pk with `from_bytes_xonly` (schnorr_verify)

theStack:
If the passed x-only-pubkey overflows (x >= p), this method fails immediately rather than silently wrapping around and then likely failing later (either because x % p is not on the curve or because the actual signature verification fails [1]). For the user of `schnorr_verify` this shouldn't make any difference, as `False` is returned with both variants, so this is probably more of a cosmetic change.

[1] `b'\xff'*32` is an example x-only-pubkey for the latter case

### DIFF
--- a/python/secp256k1proto/bip340.py
+++ b/python/secp256k1proto/bip340.py
@@ -56,7 +56,7 @@ def schnorr_verify(
     if len(sig) != 64:
         raise ValueError("The signature must be a 64-byte array.")
     try:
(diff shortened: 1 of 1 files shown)


real-or-random:
note: We may want to squash this into 843d1652e42ed3a7fb9ed9f017f414854025003f when moving secp256k1proto into a separate repo (because the latter will create a new history anyway).

PR #81, opened 2025-02-05
[CI test] Integrate secp256k1proto as subtree (using ugly import hack)

theStack:
This is a first approach of moving the `secp256k1proto` folder to an external package (living at https://github.com/theStack/secp256k1proto-playground currently, created with the uv package manager) and integrating it as git-subtree here [1]:

`$ git subtree add --prefix=python/secp256k1proto --squash https://github.com/theStack/secp256k1proto-playground master`

The import rename hack in the chilldkg_ref files to reflect the directory structure is ugly. Searching for alternatives currently.

[1] proposed solution 1 in https://groups.google.com/g/bitcoindev/c/HE3HSnGTpoQ

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -11,10 +11,10 @@ their arguments and return values, and the exceptions they raise; see also the
 from secrets import token_bytes as random_bytes
 from typing import Any, Tuple, List, NamedTuple, NewType, Optional, NoReturn, Dict
 
-from secp256k1proto.secp256k1 import Scalar, GE
-from secp256k1proto.bip340 import schnorr_sign, schnorr_verify
-from secp256k1proto.keys import pubkey_gen_plain
-from secp256k1proto.util import bytes_from_int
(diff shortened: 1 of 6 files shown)


real-or-random:
Nice!

> The import rename hack in the chilldkg_ref files to reflect the directory structure is ugly. Searching for alternatives currently.

I think you want something like this: https://github.com/BlockstreamResearch/bip-frost-dkg/tree/syspath, see https://docs.python.org/3/tutorial/modules.html#the-module-search-path for background. It's a bit low-level, but it does the job. 

This makes it possible to execute these files without a package manager. Note that `chilldkg_ref` is still a python "package". It's just managed manually.  

It may still make sense to  switch to a proper package manager in the future, as long as we make sure that you can easily execute the example and the tests. I think uv makes this somewhat easy already, but it should also be possible to make it work without uv at all. Perhaps we could [detect if we're running inside uv](https://github.com/astral-sh/uv/issues/8775) and make the `sys.path` modification conditional on that. Anyway, that's future stuff.

Issue #24, opened 2024-06-13
Extract secp256k1 code into a python package


real-or-random:
"secp256k1-proto" sounds good to me

real-or-random:
It's not yet entirely clear to me how this should like in the end. 
 - Will we include the full code of the package in the BIP? Or just refer to it, possibly pinned to some commit. The latter sounds cleaner.
 -  The package should at least have docstrings for all objects we `import`, to clarify the operations we rely on (like in BIP340, where we define stuff like `lift_x`). But then, if we don't include the full contents of the package, then the BIP is technically speaking not fully self-contained... Is that a problem? In theory, we could also run pydoc on the package and include the output in the BIP. But that sounds a bit like overkill.

real-or-random:
We discussed this somewhere else.

One additional insight is that we could also have a git subtree with the package. That would make the BIP self-contained (i.e., no external resources in external repos).

We think it's best to discuss this with the BIP editors before we take a decision.

real-or-random:
@jonasnick in https://github.com/BlockstreamResearch/bip-frost-dkg/pull/52#discussion_r1858664542:

> We discussed offline that requiring an external dependency is fine if secp256k1proto becomes a package. For now, secp256k1proto could also be a submodule instead of a package. Whatever is easier until we have a final deisionon this.

PR #82, opened 2025-02-19
python: rename secp256k1proto -> secp256k1lab

theStack:
This is a simple preparatory PR for separating the python secp256k1 code into its own repository / python package (#24). It was agreed in the working group to change the name to `secp256k1lab`, and its likely easier to reason about actual upcoming changes if the renaming and actual separation doesn't happen in the same PR.

The commit was created with the following two commands:

```
$ git mv python/secp256k1proto python/secp256k1lab
$ sed -i s/secp256k1proto/secp256k1lab/g $(git ls-files)
```


jonasnick (approved):
ACK cb8fc110dee1821576fc7746e184de7dc02f36d0

PR #83, opened 2025-02-19
Integrate secp256k1lab v1.0.0 as subtree

theStack:
This PR moves the secp256k1lab (formerly called secp256k1proto) folder to an external package (living at https://github.com/theStack/secp256k1lab currently, created with the uv package manager), integrating it as git-subtree here [1]:

```
$ git subtree add --prefix=python/secp256k1lab --squash https://github.com/secp256k1lab/secp256k1lab master
```
Supersedes #81, using the suggested syspath extension approach [1] [2] [3] to avoid the ugly import hack that was used before.

[1] https://github.com/BlockstreamResearch/bip-frost-dkg/pull/81#issuecomment-2639734301
[2] https://github.com/BlockstreamResearch/bip-frost-dkg/pull/81#issuecomment-2639741438
[3] https://github.com/BlockstreamResearch/bip-frost-dkg/tree/syspath


python/tests.py
```diff
@@ -21,6 +18,11 @@
 import chilldkg_ref.encpedpop as encpedpop
 import chilldkg_ref.chilldkg as chilldkg
 
+# Import from secp256k1lab after the chilldkg_ref imports because the latter
+# modifies sys.path to make sure the vendored copy of secp256k1proto is found.
```
jonasnick, line 22:
```suggestion
# Import from secp256k1lab after the chilldkg_ref imports because the latter
# modifies sys.path to make sure the vendored copy of secp256k1lab is found.
```

python/mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
+mypy_path = $MYPY_CONFIG_FILE_DIR/secp256k1lab/src
```
jonasnick, line 2: Can you add a comment what this does?

python/chilldkg_ref/__init__.py
```diff
@@ -1 +1,8 @@
+import os
+import sys
+
+
 __all__ = ["chilldkg"]
+
+# Prefer the vendored copy of secp256k1lab.
+sys.path.append(os.path.join(os.path.dirname(__file__), "../secp256k1lab/src"))
```
jonasnick, line 8: According to ChatGPT, this does not actually *prefer* the vendored copy because when importing python searches through the sys.path list in order. So the AI suggests

```
sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../secp256k1lab/src"))
```

jonasnick (commented):
This is looking good. Added some comments.

python/chilldkg_ref/__init__.py
```diff
@@ -1 +1,8 @@
+import os
+import sys
+
+
 __all__ = ["chilldkg"]
+
+# Prefer the vendored copy of secp256k1lab.
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../secp256k1lab/src"))
```
real-or-random, line 8: This is a bit more modern Python and also a bit more readable IMO:
```suggestion
# Prefer the vendored copy of secp256k1lab.
sys.path.insert(0, Path(__file__).parent / "../secp256k1lab/src")
```

(needs `from pathlib import Path`)

python/secp256k1lab/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "secp256k1lab"
+version = "0.1.0"
+description = "A library for prototypes using the secp256k1 curve"
```
real-or-random, line 4:
```suggestion
description = "An INSECURE implementation of the secp256k1 elliptic curve and related cryptographic schemes, intended for prototyping and education"
```

python/secp256k1lab/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+readme = "README.md"
+authors = [
+    { name = "Tim Ruffing", email = "crypto@timruffing.de" },
+    { name = "Jonas Nick", email = "jonasd.nick@gmail.com" },
+    { name = "Sebastian Falbesoner", email = "sebastian.falbesoner@gmail.com" }
+]
```
real-or-random, line 10: I think it will be fair to list @sipa as an author:
```suggestion
authors = [
    { name = "Pieter Wuille", email = "pieter@wuille.net" }, 
    { name = "Tim Ruffing", email = "me@real-or-random.org" },
    { name = "Jonas Nick", email = "jonasd.nick@gmail.com" },
    { name = "Sebastian Falbesoner", email = "sebastian.falbesoner@gmail.com" }
]
maintainers = [
    { name = "Tim Ruffing", email = "me@real-or-random.org" },
    { name = "Jonas Nick", email = "jonasd.nick@gmail.com" },
    { name = "Sebastian Falbesoner", email = "sebastian.falbesoner@gmail.com" }
]
```

based on:
https://github.com/bitcoin/bitcoin/commits/e486597f9a57903600656fb5106858941885852f/test/functional/test_framework/crypto/secp256k1.py

python/secp256k1lab/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[project]
```
real-or-random, line 1: Some additional suggestions, mostly up to debate (see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-your-pyproject-toml)



```
requires-python = ">=XXX" ## What's the minimum working version? Would be nice if 3.9 works but I can't test right now.
license = "MIT"
license-files = ["COPYING"]
keywords = ["secp256k1", "elliptic curves", "cryptography", "Bitcoin"]
classifiers = [
  "Development Status :: 5 - Production/Stable",
  "Intended Audience :: Developers",
  "Intended Audience :: Education",
  "Intended Audience :: Science/Research",
  "License :: OSI Approved :: MIT License",
  "Programming Language :: Python",
  "Topic :: Security :: Cryptography",
]

File: python/secp256k1lab/pyproject.toml [hunk 8]
```diff
@@ -0,0 +1,16 @@
...
+name = "secp256k1lab"
+version = "0.1.0"
```
real-or-random, line 3: I think we can safely make this "production", otherwise it'll be in 0.x.x land forever.

Perhaps a date-based versioning scheme like `YYYY.M.D` is better? I'd be hesitant to use `MAJOR.MINOR.REVISON` as people then think "semantic versioning" and I don't think we want to make any guarantees. This a primarily a playground. If someone wants to use it, they can vendor/pin the commit?

@jonasnick @theStack What do you think?

python/secp256k1lab/src/secp256k1lab/COPYING
real-or-random, line 1: I think we should also add

```
Copyright (c) 2025- The secp256k1lab Developers
```

real-or-random:
> I btw thought it might be nice to squash the first three commits into one, but somehow always the rename of the vendored package (src/secp256k1lab => python/secp256k1lab/src/secp256k1lab, apparently happening in the merge commit) gets lost 🤔

I think I've seen this before. Just a shot in the dark but perhaps this helps: https://stackoverflow.com/questions/30136558/how-to-squash-commits-which-have-merge-commit-in-between

python/secp256k1lab/pyproject.toml
jonasnick: lgtm

python/secp256k1lab/pyproject.toml
jonasnick: I think semantic versioning is a good choice in our case. I don't really see the problem with trying to make guarantees. We can just be very liberal about increasing the major version. Even if users vendor or pin the commit, it's very helpful to see very quickly what has changed in the library.

I agree that starting with 1.0.0 makes sense here.

jonasnick: I think that'd be helpful (at least once we have actual changes).

Are there any changes we already know we want to do before 1.0.0?

real-or-random: I won't really matter for us, but hatchling is totally fine and the uv devs choose it for a reason. https://github.com/astral-sh/uv/issues/7132. Fwiw, uv will soon have its own build backend. I'd just keep this as is.

python/secp256k1lab/pyproject.toml
real-or-random: You've added `authors` but not `maintainers`. Was this intentional? (I also don't like the duplication too much, but well, it's correct...)

python/chilldkg_ref/__init__.py
real-or-random: Oh, indeed... Well, fine, I guess it's still a tiny bit more readable than the `os.path` variants.

real-or-random:
> I guess with the ruff and mypy integration I can pretty much reuse the CI config here with minor adaptions, or is there anything special to be considered if it's a (uv) package?

This is not strictly necessary, but since we use `uv` anyway: we won't need `pip install foo`, and we can just use `uvx foo` (see `uv help tool`).

> plus determining a reasonable required Python version)

Something like `uv run --python 3.9` should do it. But you'll need to execute the BIP code: `uv run --python 3.9 python/tests.sh` (or something like this, I haven't tested.) The thing is that we can't really run the library, and there are no tests, so the best test we have is running the BIP code...

real-or-random:
Great! I didn't have a proper look but two quick nits:
 - The COPYING file should be in the repo root.
 - Can you change my email address in the toml file to me@real-or-random.org? I recently switched to that one for Bitcoin development. (Sorry, had I missed myself this when I suggested the changes...)

real-or-random:
> * add at least a very basic test first (I used this one locally: [theStack/secp256k1lab@`add_smoke_test`/smoke_test.py](https://github.com/theStack/secp256k1lab/blob/add_smoke_test/smoke_test.py?rgh-link-date=2025-03-11T01%3A48%3A27Z)); could still be added later though

I think this is good. Another good candidate for a test are the BIP340 test vectors (see https://github.com/bitcoin/bips/blob/master/bip-0340/reference.py#L147) but let's add this later then.

jonasnick (commented):
Changelog says 1.0 while pyproject.toml says 0.1. Otherwise, looks good to me.

real-or-random:
> The "crypto@" address is still used in the author metadata of your commits though, happy to change those if you want to.

This would be nice! It's really a nit but since a fresh project, it's nice to have it "clean".

PR #85, opened 2025-03-17
tests: exercise taproot tweak

wydengyre:
Augments test_vss_correctness to exercise the invalid_taproot_commit() code.


jonasnick (commented):
Thanks @wydengyre. This makes sense.

Our linter is complaining (contained in the `all.sh` script). I'll fix it in a follow-up commit.

Issue #35, opened 2024-07-08
Test vectors


jonasnick:
Add test vectors for
- [ ] correct DKG protocol runs
- [ ] incorrect contributions covering all edge cases

real-or-random:
depends on #25

siv2r:
I’d like to help create test vectors for ChillDKG’s public APIs. Here’s my current thinking:

We need test vectors for all the public functions: `hostpubkey_gen`, `params_id`, `participant_step1`, `participant_step2`, `participant_finalize`, `participant_blame`, `coordinator_step1`, `coordinator_finalize`, `coordinator_blame`, and `recover`.

Some outputs, like `ParticipantMsg1`, `ParticipantMsg2`, `CoordinatorMsg1`, `CoordinatorMsg2`, and `CoordinatorBlameMsg`, must be serialized into byte arrays since they’re transmitted between coordinator and participant. Other outputs that represent states can remain as it is?

**How many JSON files?**

My initial plan is one JSON file per public API. If that feels too fragmented later, we can merge some APIs into a single file. Starting with one file per API is more straightforward and gives us the flexibility to combine them later if needed.

**Design Options for Representing Objects**

We have two main designs for our custom types in the JSON:

1. **Design 1**: Represent it as a single large hex string (a fully serialized byte array).
2. **Design 2**: Use a nested JSON structure that shows all internal fields.

**Trade-offs**:

- Readability:
    
    Design 2 is more readable and transparent about internal structures. But will implementers actually care about this readability? Won’t they just copy and paste the test vectors into their implementation to check if it works?
    
- Complexity:
    
    With Design 2, any code using these test vectors must serialize the internal fields to match the function’s actual input/output formats. This might not be an issue, though, since developers implementing our custom type in any language will likely have serialization and deserialization functions for it.
    
- Duplication:
    
    Some data is duplicated across internal fields. For instance, in `ParticipantState1`
    
    - `t` is included in both `params` and `simpl_state`.
    - The `pubkeys` list in `params` is identical to the `enckeys` list in `enc_state`.
    - `idx` is repeated in `params`, `enc_state`, and `simpl_state`.
    
    Design 2 could eliminate this duplication by leaving it to the parsing code to infer these elements when creating a `ParticipantState1` object. This might also be possible in Design 1 by handling it within its serialization/deserialization function.
    

---

**An Example (Design 2)**

[...]

PR #90, opened 2025-03-30
Add JSON Test Vectors

siv2r:
Addresses issue #35.

This PR introduces functionality for generating and validating JSON test vectors for ChillDKG’s public APIs.

Summary:

- Implements [Design 3](https://github.com/BlockstreamResearch/bip-frost-dkg/issues/35#issuecomment-2565855499), with assertions for [intermediate messages](https://github.com/BlockstreamResearch/bip-frost-dkg/issues/35#issuecomment-2577725803).
- Modules in `python/vectors_generator/` generate one JSON vector file per API, saving them in the `python/vectors/`.
- Modifies `chilldkg_ref`:
    1. To ensure reproducible test vectors, this PR temporarily uses a zeroed `aux_rand` value for `schnorr_sign`
    2. Adds the `GE.from_bytes_with_infinity` method.
- Adds tests within `tests.py` to load the generated JSON vectors and verify API outputs for correctness.


jonasnick (commented):
Nice, thank you! This looks like "design 3" as I had it in mind.

I was surprised, shouldn't this be based on #88? I think the test vectors should contain the messages in their serialized form.

python/vector_generator/coordinator.py
```diff
@@ -0,0 +1,247 @@
...
+        chilldkg.FaultyParticipantError,
+    )
+    vectors["error_test_cases"].append(
+        {
+            "pmsgs1": [pmsg1_asdict(m) for m in pmsgs1],
+            "params": params_asdict(params),
+            "cmsg1": cmsg1_asdict(cmsg1),
+            "pmsgs2": [pmsg2_asdict(m) for m in invalid_pmsgs2],
+            "error": error,
+            "comment": "invalid certificate length",
```
jonasnick, line 206: wrong comment

python/vector_generator/participant.py
```diff
@@ -0,0 +1,532 @@
...
+    )
+    vectors["error_test_cases"].append(
+        {
+            "hostseckey": bytes_to_hex(hostseckeys[0]),
+            "params": params_asdict(params),
+            "random": bytes_to_hex(randoms[0]),
+            "pmsg1": pmsg1_asdict(pmsgs1[0]),
+            "cmsg1": cmsg1_asdict(invalid_cmsg1),
+            "error": error,
+            "comment": "coms_to_secret list in cmsg1 has an invalid value at index 0",
```
jonasnick, line 248: wrong comment

python/vector_generator/participant.py
```diff
@@ -0,0 +1,532 @@
...
```
jonasnick, line 268: wrong comment

python/vector_generator/others.py
```diff
@@ -0,0 +1,352 @@
...
+)
+
+from chilldkg_ref.chilldkg import hostpubkey_gen, params_id, recover
+import chilldkg_ref.chilldkg as chilldkg
+
+
+def generate_hostpubkey_vectors():
+    vectors = {"valid_test_cases": [], "error_test_cases": []}
+
+    # --- Valid test case 1 ---
```
jonasnick, line 18: nit: can we start numbering with 0? The test cases are stored in a list, where the conventional first index is 0.

jonasnick (commented):
Posting a status update, still reviewing...

siv2r:
Thanks for the review. I'll address your comments soon. 

> I was surprised, shouldn't this be based on https://github.com/BlockstreamResearch/bip-frost-dkg/pull/88? I think the test vectors should contain the messages in their serialized form.

I assumed we were favoring design 2 over design 1. That is why the JSON representation of messages shows internal fields (for better readability) rather than a single byte array. The definitions of design 1 and design 2 are in [this comment](https://github.com/BlockstreamResearch/bip-frost-dkg/issues/35#issuecomment-2553866102) as:

> We have two main designs for our custom types in the JSON:
> 
> 1. **Design 1**: Represent it as a single large hex string (a fully serialized byte array).
> 2. **Design 2**: Use a nested JSON structure that shows all internal fields.

I can switch the message JSON to a single hex-encoded byte array if you prefer.

python/vector_generator/others.py
```diff
@@ -0,0 +1,352 @@
...
+            "error": error,
+            "comment": "first coefficient of sum_coms is invalid",
+        }
+    )
+    # --- Error test case 3: last share in enc_secshare list is out of range ---
+    n = len(hostpubkeys)
+    cert_len = chilldkg.certeq_cert_len(n)
+    rand_encshare = bytes.fromhex(
+        "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141"
+    )
```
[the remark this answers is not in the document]
siv2r: Ah yes, thanks.

jonasnick:
I had thought design 1 vs. design 2 was mostly about representing the state objects. As for the messages that are passed between the participants, I believe it's better to include them in the test vectors in their serialized form. This way, the byte encoding is tested as well, which is an important part of the spec.

python/tests.py
```diff
@@ -362,6 +381,306 @@ def test_correctness(t, n, simulate_dkg, recovery=False, investigation=False):
             assert pubshares == dkg_outputs[i][2]
 
 
+def test_hostpubkey_gen_vectors():
+    input_file = Path("vectors/hostpubkey_gen_vectors.json")
+    with open(input_file) as f:
```
jonasnick, line 386: When run from the parent directory via `python python/tests.py` this gives me an error:
```
FileNotFoundError: [Errno 2] No such file or directory: 'vectors/hostpubkey_gen_vectors.json'
```
`python python/tests.py` works fine on master

python/vector_generator/coordinator.py
```diff
@@ -0,0 +1,247 @@
...
+            "error": error,
+            "comment": "hostpubkeys list contains and invalid value",
+        }
+    )
+    # --- Error Test Case 4: Participant (index 1) message has an enc_shares list of invalid length ---
+    invalid_pmsgs1 = copy.deepcopy(pmsgs1)
+    invalid_pmsgs1[1].enc_pmsg.enc_shares.pop()
+    error = expect_exception(
+        lambda: coordinator_step1(invalid_pmsgs1, params),
+        chilldkg.FaultyParticipantOrCoordinatorError,  # REVIEW: why is this working? shouldn't it raise an error saying encpedpop.FaultyParticipantOrCoordinatorError was raised instead?
```
jonasnick, line 110: Why would it raise "`encpedpop.FaultyParticipantOrCoordinatorError`"? `FaultyParticipantOrCoordinatorError` is defined in util.py and it's part of the chilldkg module.

mllwchrry (commented):
I've identified several coverage gaps in the test vectors, aside from those mentioned by @jonasnick and the TODO statements. Consider adding cases below.

General improvements:
- Add valid test cases for different threshold settings (not just 2-of-3).
- Many fields are currently only tested at index 0 or 1; ensure logic is tested against both the caller's index and some other participant index. This is especially important for operations that validate or process data per-participant.
- Whenever `hostseckey` is used, along with existing tests, test for zero key and `hostseckey` >= group order.

Participants logic:
- In `participant_step1` t > n test is missing.
- In `participant_finalize` there's a test for an invalid certificate length with too short certificate; consider adding a check where the certificate is too long (same for `coordinator_finalize`).

Recovery logic:
- A test for duplicate `pubnonces` in the `pubnonces` list is missing.

python/gen_vector_utils/coordinator.py
```diff
@@ -0,0 +1,247 @@
...
+    participant_step1,
+    participant_step2,
+    coordinator_step1,
+    coordinator_finalize,
+    coordinator_investigate,
+)
+import chilldkg_ref.chilldkg as chilldkg
+
+
+def generate_coordinator_step1_vectors():
```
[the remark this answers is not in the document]
jonasnick: Let's defer more exhaustive test coverage to subsequent PRs.

python/gen_vector_utils/participant.py
```diff
@@ -0,0 +1,532 @@
...
+            "hostseckey": bytes_to_hex(hostseckeys[0]),
+            "params": params_asdict(params),
+            "random": bytes_to_hex(randoms[0]),
+            "pmsg1": pmsg1_asdict(pmsgs1[0]),
+            "cmsg1": cmsg1_asdict(cmsg1),
+            "expected_pmsg2": pmsg2_asdict(pmsg2),
+        }
+    )
+
+    # --- Error Test Case 1: Wrong host secret key length ---
```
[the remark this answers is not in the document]
siv2r: It should be have been removed. Fixed.

python/tests.py
siv2r: Fixed

python/vector_generator/coordinator.py
siv2r: Yes, makes sense. Thank you.

python/gen_vector_utils/coordinator.py
```diff
@@ -0,0 +1,247 @@
...
+    error = expect_exception(
+        lambda: coordinator_step1(pmsgs1, invalid_params),
+        chilldkg.InvalidHostPubkeyError,
+    )
+    vectors["error_test_cases"].append(
+        {
+            "pmsgs1": [pmsg1_asdict(m) for m in pmsgs1],
+            "params": params_asdict(invalid_params),
+            "error": error,
+            "comment": "hostpubkeys list contains and invalid value",
```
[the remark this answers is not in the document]
siv2r: fixed

python/gen_vector_utils/participant.py
```diff
@@ -0,0 +1,532 @@
...
+    hostpubkeys = [chilldkg.hostpubkey_gen(sk) for sk in hostseckeys]
+    params = chilldkg.SessionParams(hostpubkeys, 2)
+    randoms = hex_list_to_bytes(
+        [
+            "42B53D62E27380D6F7096EDA1C28C57DDB89FCD4CE5B843EDAC220E165B5A7EC",
+            "FDE223740111491D5E60BEFB447A2D8C0B12D4B1CE1A0D6BF5A16CBA7E420153",
+            "E5CFC54DA8EE57BA97C389060D00BB840A9DDF6BF1E32AE3D3598373EF384EE7",
+        ]
+    )
+    assert len(randoms) == len(hostpubkeys)
```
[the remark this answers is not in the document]
siv2r: Ah yes, I kept this check in case these two lists are modified incorrectly in the future. They need to stay coupled. I think it's safe to keep it?

siv2r:
@jonasnick @mllwchrry, this PR is ready for another round of review!

Since the last review, a couple of things have changed upstream (message ser/deser merged, `aux_rand` parameter added to `participant_step2`). On top of that, here's what's new:
- Participant and coordinator messages are now encoded as serialized bytes in the JSON vectors, instead of the earlier nested-value format.
- Test vectors are fully reproducible, `aux_rand` is set to a fixed value so re-running `gen_vectors.py` always produces the same output.

I've addressed most of jonasnick's review comments:
- Reordered the error test cases to match the validation order in the code (e.g., threshold is checked before duplicate hostpubkeys).
- Added a test for wrong randomness length in `participant_step1`.
- Added a test for tampered `sum_coms_to_nonconst_terms` in `participant_step2`.
- Fixed `FileNotFoundError` when running `python python/tests.py` from the repo root.
- Various other small fixes (comments, naming, documentation).

I also brought in a few ideas from the wycheproof test vector format: specifically `tc_id`, `description`, and `total_tests` fields. The bigger wycheproof patterns (test groups referencing shared fields, valid/invalid/acceptable tags) weren't needed since we already have shared fields at the top of each JSON file and separate lists for valid and error test cases.

These vectors aren't exhaustive. This PR establishes the structure and baseline for how ChillDKG test vectors will look. More thorough coverage will come in follow-up PRs.

siv2r:
One more thing, regarding testing different `(t, n)` configurations (like mllwchrry mentioned): we could adopt wycheproof's `testGroups` pattern for this. Each group would carry its own `params`, shared setup fields, and test cases. That way, we can test multiple configurations (e.g., standard `n=3, t=2` alongside degenerate `n=1, t=1`) in a single file while keeping `tc_id` unique across groups. Something to consider for a follow-up PR.

```json                                                                                                                                                      
{                                                                                                                                                            
  "testGroups": [                                                                                                                                            
    {                                                                                                                                                        
      "description": "Standard 2-of-3 DKG",                                                                                                                  
      "params": { "hostpubkeys": [...], "t": 2 },                                                                                                            
      "hostseckey": "ADE1...",                                                                                                                               
      "random": "42B5...",                                                                                                                                   
      "aux_rand": "005F...",                                                                                                                                 
      "pmsg1": "036B...",                                                                                                                                    
      "valid_test_cases": [...],                                                                                                                             
      "error_test_cases": [...]                                                                                                                              
    },                                                                                                                                                       
    {                                                                                                                                                        
      "description": "Degenerate 1-of-1 DKG",                                                                                                                
      "params": { "hostpubkeys": ["03AB..."], "t": 1 },                                                                                                      
      "hostseckey": "...",                                                                                                                                   
      "random": "...",                                                                                                                                       
      "aux_rand": "...",                                                                                                                                     
      "pmsg1": "...",                                                                                                                                        
      "valid_test_cases": [...],                                                                                                                             
      "error_test_cases": [...]                                                                                                                              
    }                                                                                                                                                        
  ]                                                                                                                                                          
}                                                                                                                                                            
```

siv2r:
Our pool arrays (`pmsg1_pool`, `cmsg1_pool`, etc.) are just lists of hex blobs with no metadata. When a test case references `pmsg1_indices: [0, 1, 3]`, you have to guess what pool entry 3 is.

We could add a parallel `_annotations` array next to each pool, one string per entry describing what it is and, for invalid entries, what was changed:

```json
{
  "pmsg1_pool": [
    "036BEE122E...",
    "037C42E200...",
    "026F714517...",
    "037C42E200...truncated"
  ],
  "pmsg1_pool_annotations": [
    "valid pmsg1 from participant 0",
    "valid pmsg1 from participant 1",
    "valid pmsg1 from participant 2",
    "participant 1's pmsg1 truncated by 32 bytes (missing encrypted shares)."
  ]
}
```

The generator script already knows what it's constructing for each entry, so this is basically free to maintain. Would like to know if you think it's worth including this extra `_annotations` field.

mllwchrry:
The PR is good now. Found only a few typos and one suggestion:

1. Typo in `participant.py`. Lines 270, 286, 289, 303 have comments referring to `coms_to_secret` but the actual field name is `coms_to_secrets`.
2. The same test data (`hostseckeys`, `randoms`, `aux_rand`, invalid signatures, etc.) is duplicated across different files for generating vectors. These could be extracted into a separate file with constants, which would make maintenance easier. However, this could be done in another PR.

I also like the ideas you proposed for `testGroups` and `_annotations`. They are great enhancements!

python/gen_vector_utils/participant.py
```diff
@@ -0,0 +1,532 @@
...
+    vectors["error_test_cases"].append(
+        {
+            "hostseckey": bytes_to_hex(rand_hostseckey),
+            "params": params_asdict(params),
+            "random": bytes_to_hex(random),
+            "error": error,
+            "comment": "host secret key doesn't match any hostpubkey",
+        }
+    )
+    # --- Error test case 3: Invalid threshold ---
```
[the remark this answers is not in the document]
siv2r: Fixed

python/gen_vector_utils/participant.py
```diff
@@ -0,0 +1,532 @@
...
+            "random": bytes_to_hex(random),
+            "error": error,
+            "comment": "hostpubkeys list contains an invalid value",
+        }
+    )
+
+    return vectors
+
+
+def generate_participant_step2_vectors():
```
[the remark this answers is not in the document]
siv2r: Done

jonasnick:
The code looks good and implements the design we discussed. Let's defer solving the remaining issues to follow-up PRs. Thanks @siv2r and @mllwchrry!

PR #100, opened 2026-01-30
Make *MsgParseError internal

siv2r:
As suggested in https://github.com/BlockstreamResearch/bip-frost-dkg/pull/88#issuecomment-3822647429, this pull request makes the `ParticipantMsgParseError` and `CoordinatorMsgParseError` internal. Users catch these exceptions as `FaultyParticipantError` and `FaultyCoordinatorError` respectively.

### DIFF
--- a/python/chilldkg_ref/util.py
+++ b/python/chilldkg_ref/util.py
@@ -104,20 +104,12 @@ class UnknownFaultyParticipantOrCoordinatorError(ProtocolError):
 
 
 class MsgParseError(ValueError):
-    """Raised when parsing a serialized message fails."""
+    pass
 
 
-class ParticipantMsgParseError(MsgParseError):
-    """Raised when parsing a participant message fails.
-
-    Attributes:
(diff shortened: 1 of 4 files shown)


python/chilldkg_ref/util.py
```diff
@@ -104,20 +104,13 @@ def __init__(self, inv_data: Any, *args: Any):
...
-    """Raised when parsing a participant message fails.
-
-    Attributes:
-        participant (int): Index of the participant whose message failed to parse.
-    """
-
+class ParticipantMsgParseError(FaultyParticipantError):
     def __init__(self, participant: int, *args: Any):
-        self.participant = participant
         super().__init__(participant, *args)
```
jonasnick: I believe we can just use `pass` here as well.

siv2r: Yes. Fixed it.

real-or-random (commented):
Post-merge ACK

Issue #31, opened 2024-07-03
Domain Separation for PoPs is insufficient


jonasnick:
While thinking about https://github.com/BlockstreamResearch/bip-frost-dkg/issues/30#issuecomment-2202630082, I came up with an attack that I'm pretty sure works against the current ChillDKG implementation. The attack requires "deterministic randomness" and the absence of proper domain separation between PoPs and FROST signatures.

The main idea of the attack is as follows: Assume the victim contributes randomness `x1` via `X1 = x1*G` in every session. The adversary uses a signature from the victim in session 1 as PoPs in session 2 which allows him to contribute randomness `X1 + Y`. By continuing to use signatures in session i as PoPs in session i+1 the adversary is able to double the contributed randomness and obtain the corresponding PoPs from the victim until the adversary has a PoP for `(n-1)*X1 + t*Y = -X1 + t*Y` which allows key cancellation.

In more detail:
1. In every session the victim deterministically contributes randomness `x1` and `X1 = x1*G`. Assume wlog that victim is at signer index 1.
2. The adversary requests 2-of-2 DKG session with the victim, providing some randomness `1*G = G`. The threshold public key that is output of the DKG is `X1 + G`.
3. The adversary requests two FROST signatures from the victim with the messages chosen such that the signature will serve as a PoP for `X1 + G` and signer indices 2 and 3 in the subsequent session. The adversary completes the signature.
4. The adversary requests 3-of-3 DKG session with the victim. This time the adversary provides randomness `X1 + G` using the PoP obtained through regular FROST signing in the previous step. The adversary correspondingly sends VSS commitment `(X1 + G, X2', X3')` where `X2'` and `X3'` are chosen such that `f(1) = 2*G = X1 + G + X2' + X3'` where which makes `2` a valid share for the victim signer at index 1. The adversary sends the share to the victim.
5. This concludes the DKG: the threshold public key is `3*X1 + 2*G` and the victim and adversary have all the shares necessaries to sign for it.
6. The adversary requests a signature from the victim and completes it to a signature for `3*X1 + 2*G` with the messages chosen such that the signature will serve as a PoP for `X1 + G` and signer indices 2 and 3 in the subsequent session.
7. Adversary requests 3-of-3 DKG session with the victim but now provides VSS commitment `(3*X1 + 2*G, X2', X3')` and uses the signature from the previous step for the PoP of `3*X1 + 2*G`.
[...]

PR #101, opened 2026-02-03
Expose `aux_rand` to ChillDKG APIs for derterministic behaviour

### LINKED ISSUE #86: Non-Reproducible Test Vectors due to Fresh `aux_rand`
I'm implementing test vectors for ChillDKG. To ensure reproducibility, I use fixed values for `random` and `hostseckey` inputs, so every run of the test vector generation script yields the same vectors.

However, two internal functions, `certeq_participant_step` and `pop_prove`, currently call `schnorr_sign` using a freshly generated random 32-byte value for the `aux_rand` parameter. This fresh randomness prevents the consistent reproduction of test vectors.

https://github.com/BlockstreamResearch/bip-frost-dkg/blob/1e34161e1ce5c1042015373adc1bb0dfa5d4c4b7/python/chilldkg_ref/chilldkg.py#L81-L83

https://github.com/BlockstreamResearch/bip-frost-dkg/blob/1e34161e1ce5c1042015373adc1bb0dfa5d4c4b7/python/chilldkg_ref/simplpedpop.py#L38-L41

Specifically:

- `participant_step1` produces varying outputs (`pmsg1`) between runs because the `Pop` component generated by `pop_prove` differs each time, even with fixed inputs.
- Similarly, `participant_step2` outputs vary because of the randomness introduced by `certeq_participant_step`.

These output variations propagate, affecting other participant and coordinator APIs relying on `participant_step1` and `participant_step2`.

**Question:** Is it a good idea to replace the fresh randomness with deterministically derived values (via a tagged hash) to achieve reproducibility?

siv2r:
Fixes #86 

Modified ChillDKG APIs to accept randomness parameters rather than generating new randomness internally. This enables deterministic behavior when needed (e.g., for test vectors).                                                                                                             

- Add `random` parameter to `pop_prove()` and `certeq_participant_step()`
- Add `random` parameter to `simplpedpop.participant_step1()` and `chilldkg.participant_step2()`
- Update tests and example code to use new APIs 
- Fix `e.args` unpacking in encpedpop

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -10,7 +10,6 @@ their arguments and return values, and the exceptions they raise; see also the
 
 from __future__ import annotations
 
-from secrets import token_bytes as random_bytes
 from typing import Any, Tuple, List, NamedTuple, NewType, Optional, NoReturn, Dict
 
 from secp256k1lab.secp256k1 import Scalar, GE
@@ -88,9 +87,11 @@ def certeq_message(x: bytes, idx: int) -> bytes:
     return prefix + idx.to_bytes(4, "big") + x
 
 
-def certeq_participant_step(hostseckey: bytes, idx: int, x: bytes) -> bytes:
+def certeq_participant_step(
+    hostseckey: bytes, idx: int, x: bytes, aux_rand: bytes
+) -> bytes:
     msg = certeq_message(x, idx)
-    return schnorr_sign(msg, hostseckey, aux_rand=random_bytes(32))
+    return schnorr_sign(msg, hostseckey, aux_rand=aux_rand)
 
 
 def certeq_cert_len(n: int) -> int:
@@ -577,6 +578,7 @@ def participant_step2(
     hostseckey: bytes,
     state1: ParticipantState1,
     cmsg1: bytes,
+    aux_rand: bytes,
 ) -> Tuple[ParticipantState2, bytes]:
     """Perform a participant's second step of a ChillDKG session.
 
@@ -597,6 +599,9 @@ def participant_step2(
         state1: The participant's session state as output by
             `participant_step1`.
         cmsg1: The first message received from the coordinator.
+        aux_rand: Auxiliary randomness (32 bytes). FRESH 32-byte randomness
+            is optimal, but 16 random bytes or a counter padded to 32 bytes
+            is acceptable (see BIP 340).
 
     Returns:
         ParticipantState2: The participant's session state after this step, to
@@ -607,6 +612,7 @@ def participant_step2(
 
     Raises:
         HostSeckeyError: If the length of `hostseckey` is not 32 bytes.
+        RandomnessError: If the length of `aux_rand` is not 32 bytes.
         FaultyCoordinatorError: If the coordinator is faulty. See the
             documentation of the exception for further details.
         FaultyParticipantOrCoordinatorError: If another known participant or the
@@ -621,6 +627,8 @@ def participant_step2(
     """
     if len(hostseckey) != 32:
         raise HostSeckeyError
+    if len(aux_rand) != 32:
+        raise RandomnessError
 
     params, idx, enc_state = state1
     t = enc_state.simpl_state.t
@@ -642,7 +650,7 @@ def participant_step2(
     eq_input += b"".join([bytes_from_int(int(share)) for share in enc_secshares])
     dkg_output = DKGOutput._make(enc_dkg_output)
     state2 = ParticipantState2(params, eq_input, dkg_output)
-    sig = certeq_participant_step(hostseckey, idx, eq_input)
+    sig = certeq_participant_step(hostseckey, idx, eq_input, aux_rand)
     pmsg2 = ParticipantMsg2(sig).to_bytes()
     return state2, pmsg2
 
--- a/python/chilldkg_ref/simplpedpop.py
+++ b/python/chilldkg_ref/simplpedpop.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from secrets import token_bytes as random_bytes
 from typing import List, NamedTuple, NewType, Tuple, Optional, NoReturn
 
 from secp256k1lab.bip340 import schnorr_sign, schnorr_verify
@@ -40,10 +39,8 @@ def pop_msg(idx: int) -> bytes:
     return idx.to_bytes(4, byteorder="big")
 
 
-def pop_prove(seckey: bytes, idx: int) -> Pop:
-    sig = schnorr_sign(
-        pop_msg(idx), seckey, aux_rand=random_bytes(32), tag_prefix=POP_MSG_TAG
-    )
+def pop_prove(seckey: bytes, idx: int, aux_rand: bytes) -> Pop:
+    sig = schnorr_sign(pop_msg(idx), seckey, aux_rand=aux_rand, tag_prefix=POP_MSG_TAG)
     return Pop(sig)
 
 
@@ -220,7 +217,7 @@ class ParticipantInvestigationData(NamedTuple):
 
 
 def participant_step1(
-    seed: bytes, t: int, n: int, idx: int
+    seed: bytes, t: int, n: int, idx: int, aux_rand: bytes
 ) -> Tuple[
     ParticipantState,
     bytes,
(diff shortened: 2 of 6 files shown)


python/chilldkg_ref/chilldkg.py
```diff
@@ -80,17 +79,21 @@
...
     return prefix + idx.to_bytes(4, "big") + x
 
 
-def certeq_participant_step(hostseckey: bytes, idx: int, x: bytes) -> bytes:
+def certeq_participant_step(
+    hostseckey: bytes, idx: int, x: bytes, aux_rand: bytes
+) -> bytes:
     msg = certeq_message(x, idx)
-    return schnorr_sign(msg, hostseckey, aux_rand=random_bytes(32))
+    return schnorr_sign(msg, hostseckey, aux_rand=aux_rand, tag_prefix=CERTEQ_MSG_TAG)
```
jonasnick: Adding a `tag_prefix` is a protocol breaking change that is independent of the intent of the commit as specified in the commit message ("expose aux_rand in APIs for deterministic behavior").
I'm not sure what the motivation is. We already domain separate via the message prefix. We had switched to this approach for domain separation in PR #69.

python/chilldkg_ref/chilldkg.py
```diff
@@ -642,7 +653,11 @@ def participant_step2(
     eq_input += b"".join([bytes_from_int(int(share)) for share in enc_secshares])
     dkg_output = DKGOutput._make(enc_dkg_output)
     state2 = ParticipantState2(params, eq_input, dkg_output)
-    sig = certeq_participant_step(hostseckey, idx, eq_input)
+
+    # Derive aux_rand from random for CertEq signature
+    aux_rand = tagged_hash_bip_dkg("certeq aux_rand", random)
+    sig = certeq_participant_step(hostseckey, idx, eq_input, aux_rand)
```
jonasnick: Why not pass randomness directly to certeq?

python/chilldkg_ref/simplpedpop.py
```diff
@@ -236,10 +234,14 @@ def participant_step1(
...
         raise ValueError
+    if len(random) != 32:
+        raise ValueError
 
     vss = VSS.generate(seed, t)  # OverflowError if t >= 2**32
     partial_secshares_from_me = vss.secshares(n)
-    pop = pop_prove(vss.secret().to_bytes(), idx)
+
+    aux_rand = tagged_hash_bip_dkg("pop aux_rand", random)
+    pop = pop_prove(vss.secret().to_bytes(), idx, aux_rand)
```
jonasnick: Why not pass randomness directly to pop_prove?

python/chilldkg_ref/chilldkg.py
```diff
@@ -577,8 +581,10 @@ def participant_step2(
     hostseckey: bytes,
     state1: ParticipantState1,
     cmsg1: bytes,
+    random: bytes,
 ) -> Tuple[ParticipantState2, bytes]:
-    """Perform a participant's second step of a ChillDKG session.
+    """Perform a participant's second step of a ChillDKG session. This function
+    must be called exactly once for each `participant_step1` call.
```
jonasnick: I think it's better to remove "This function must be called exactly once for each `participant_step1` call" from this PR because I believe that it may be confusing to the reader and if we want to be consistent we need to add this to other functions as well.

python/chilldkg_ref/chilldkg.py
```diff
@@ -597,6 +603,8 @@ def participant_step2(
         state1: The participant's session state as output by
             `participant_step1`.
         cmsg1: The first message received from the coordinator.
+        random: FRESH random byte string (32 bytes). **MUST** be independent
+            from the `random` value used in `participant_step1`.
```
jonasnick: "**MUST** be independent" is implied by "FRESH" (and imho sounds unnecessarily scary)

python/chilldkg_ref/chilldkg.py
siv2r: Thanks for pointing this out! I introduced `tag_prefix` here because `schnorr_sign` defaults to the "BIP340" tag when generating nonce and challenge hashes. I assumed we should use the "BIP DKG" tag instead, similar to how `pop_prove` specifies a tag in simplpedpop. I thought this might have been missed.

I didn't have context about PR 69. I'll revert this change.

python/chilldkg_ref/chilldkg.py
siv2r: Fixed in a31ab00523982980f18da81b7595178f7a5faaa7

python/chilldkg_ref/chilldkg.py
siv2r: Fixed in a31ab00523982980f18da81b7595178f7a5faaa7

python/chilldkg_ref/simplpedpop.py
siv2r: Fixed in a31ab00523982980f18da81b7595178f7a5faaa7

python/chilldkg_ref/chilldkg.py
siv2r: Fixed in a31ab00523982980f18da81b7595178f7a5faaa7

python/chilldkg_ref/chilldkg.py
siv2r: Fixed in a31ab00523982980f18da81b7595178f7a5faaa7

real-or-random: see also https://github.com/secp256k1lab/secp256k1lab/issues/20

real-or-random: ~It's not related to this PR directly, but now after opening that issue, I'm noticing that we still use a different tag for the pops. (Sorry siv, I should have read your comment above properly.)  @jonasnick I remember vaguely that we talked about this and probably had some reason to keep the separate tag for the pops, but I can't exactly remember why.~

~Perhaps I was thinking that PoPs/ZKPoKs are a different primitive than sigs, so they should be separated? If so, this is true in theory, but in hindsight, it seems pointless from an engineering point of view: Any Schnorr sig is a ZKPoK, so that direction is fine. And in the other direction, you would want to avoid mistaking a ZKPoK for a signature of some specific message, but that's precisely what domain separation solves. So this could be realized just like we do it in certeq? That would certainly make the protocol easier to implement because it works with existing BIP340 implementations then.~

python/chilldkg_ref/chilldkg.py
real-or-random: I admit that I was a bit on the fence here. Deriving `aux_rand` via the PRF is cheap and clean. And it makes it easier to add more other calls that require randomness (e.g., in future versions) because then all derivations will be properly separated. 

But ultimately, I agree it's simpler to pass `random` directly down. And in that case (and also in the case below for the pops), we only pass it to `aux_rand` of schnorrsigs, where it's totally fine: a signature won't leak anything about `aux_rand`. 

I'd prefer to call the new API argument also `aux_rand` because that's what it is in the end.  And it should be documented as auxiliary only. Nothing bad happens if the user passes in a constant value.

python/chilldkg_ref/simplpedpop.py
real-or-random: same here

real-or-random:
What are your thoughts on making the new arguments optional? If the caller does not provide them, we'll let the function call `random_bytes(32)`. This keeps the interface simple while still allowing for deterministic behavior for the test vectors.

python/chilldkg_ref/chilldkg.py
jonasnick: This was the reason: https://github.com/BlockstreamResearch/bip-frost-dkg/issues/31.

real-or-random: Nevermind, #31 is the right reason to keep pops as they are.

jonasnick:
> What are your thoughts on making the new arguments optional?

Either way is fine for me.

siv2r:
Renamed the random parameter to `aux_rand` in `certeq_participant_step`, `chilldkg.participant_step2`, `simplpedpop.participant_step1`, and `simplpedpop.pop_prove`, since in all these functions, `random` is used solely as `aux_rand` in a Schnorr signature call. Updated the docstring for `chilldkg.participant_step2` accordingly.

siv2r:
> What are your thoughts on making the new arguments optional? If the caller does not provide them, we'll let the function call `random_bytes(32)`.

Making `aux_rand` optional is nice for callers who don't care about determinism. One minor concern: a caller skipping the argument might misread it as "no additional randomness is used internally," when in fact `random_bytes(32)` would still be called (this can be easily addressed with a doc note).

Since this is a reference implementation, I lean slightly towards keeping `aux_rand` required: it's explicit about every input entering the protocol, and we don't need to worry about caller sophistication. That said, I'm not strongly opinionated here. Happy to make it optional if you prefer that.

jonasnick (approved):
ACK efa493d30aed5c9dfb8abecbce3ba998981b2c28

PR #106, opened 2026-04-02
Add recovery acknowledgment functions

### LINKED ISSUE #36: Implement explicit "confirmation" messages
An explicit spec is important, we should not be lazy and leave this to implementers.

mllwchrry:
Addresses #36.

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -44,6 +44,8 @@ __all__ = [
     "coordinator_finalize",
     "coordinator_investigate",
     "recover",
+    "participant_recovery_ack_sign",
+    "participant_recovery_acks_verify",
     # Exceptions
     "HostSeckeyError",
     "SessionParamsError",
@@ -57,6 +59,7 @@ __all__ = [
     "FaultyCoordinatorError",
     "UnknownFaultyParticipantOrCoordinatorError",
     "RecoveryDataError",
+    "InvalidRecoveryAckError",
     # Types
     "SessionParams",
     "DKGOutput",
@@ -119,6 +122,25 @@ class InvalidSignatureInCertificateError(ValueError):
     def __init__(self, participant: int, *args: Any):
         self.participant = participant
         super().__init__(participant, *args)
+
+
+###
+### Recovery acknowledgment helpers
+###
+
+
+def recovery_ack_message(x: bytes, idx: int) -> bytes:
+    # Domain separation as described in BIP 340
+    prefix = (BIP_TAG + "recovery acknowledgment").encode()
+    prefix = prefix + b"\x00" * (33 - len(prefix))
+    return prefix + idx.to_bytes(4, "big") + x
+
+
+def recovery_ack_sign(
+    hostseckey: bytes, idx: int, x: bytes, aux_rand: bytes
+) -> bytes:
+    msg = recovery_ack_message(x, idx)
+    return schnorr_sign(msg, hostseckey, aux_rand=aux_rand)
 
 
 ###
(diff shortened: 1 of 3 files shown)


mllwchrry:
Currently, the code raises `ValueError` in the new public function `participant_recovery_acks_verify`, similar to how we do it for length checks in other public functions (`coordinator_step1` and `coordinator_finalize`). However, this exception is never documented in the spec for any of these functions and seems to be non-informative.
Should we open a separate PR to address this? We could either document `ValueError` in the Raises sections of all three functions, or create more specific exceptions.

real-or-random:
Concept ACK but this needs rebase

mllwchrry:
Rebased, no conflicts now

python/chilldkg_ref/chilldkg.py
```diff
@@ -121,6 +124,25 @@ def __init__(self, participant: int, *args: Any):
...
 
+###
+### Recovery acknowledgment helpers
+###
+
+
+def recovery_ack_message(x: bytes, idx: int) -> bytes:
+    # Domain separation as described in BIP 340
+    prefix = (BIP_TAG + "recovery acknowledgment").encode()
+    prefix = prefix + b"\x00" * (33 - len(prefix))
```
real-or-random, line 136: Hm, this code is not robust against changing `recovery acknowledgment`. If you make that string too long, it may happen currently does not guarantee that the final prefix will be 33 byte. Perhaps add an assertion? 

But I see that you simply took this from certeq, so this is where we made the mistake initially. If we add an assertion here, we should also add one in certeq.

python/chilldkg_ref/chilldkg.py
```diff
@@ -998,3 +1020,134 @@ def recover(
...
+
+
+def participant_recovery_acks_verify(
+    recovery_data: RecoveryData, params: SessionParams, ack_sigs: List[bytes]
+) -> None:
+    """Verify recovery acknowledgment signatures from all participants.
+
+    This function is used to ensure that all participants have
+    received the recovery data before the threshold public key is used
+    (e.g., before funds are sent to it).
```
real-or-random, line 1120: I think the wording can be made more consistent here and in the BIP text.

We currently use "acknowledgment" and "confirmation" interchangeably. It will be clearer to use only once. I'm fine with either but I'm leaning towards "acknowledgment". I think it's more common when it comes to receiving messages.

python/chilldkg_ref/chilldkg.py
```diff
@@ -998,3 +1020,134 @@ def recover(
...
+        valid = schnorr_verify(
+            msg,
+            hostpubkeys[i][1:33],
+            sig,
+        )
+        if not valid:
+            raise InvalidRecoveryAckError(i)
+
+
+class InvalidRecoveryAckError(ValueError):
```
real-or-random: Should this subclass `FaultyParticipantError`? I think so.

python/chilldkg_ref/chilldkg.py
```diff
@@ -998,3 +1020,134 @@ def recover(
...
+            from all participants, in the same order as `hostpubkeys`.
+
+    Raises:
+        InvalidHostPubkeyError: If `hostpubkeys` contains an invalid public key.
+        DuplicateHostPubkeyError: If `hostpubkeys` contains duplicates.
+        ThresholdOrCountError: If `1 <= t <= len(hostpubkeys) <= 2**32 - 1` does
+            not hold.
+        RecoveryDataError: If the recovery data is invalid or does not match
+            the provided parameters.
+        InvalidRecoveryAckError: If any recovery acknowledgment signature is
```
real-or-random, line 1135: It's a bit subtle: If you get `InvalidRecoveryAckError`, this doesn't mean that the DKG session has failed. (On the contrary: if you reach this state, then you know that the DKG was "successful", at least according to our definition of successful). You just don't know that everyone has the recovery data.

mllwchrry:
Addressed review comments and added BIP clarifications from our discussion. @real-or-random, ready for re-review.

python/chilldkg_ref/chilldkg.py
```diff
@@ -665,6 +689,11 @@ def participant_finalize(
     the session successful by presenting the recovery data to them, from which
     they can recover the DKG outputs using the `recover` function.
 
+    Returning successfully only means the DKG is complete. It does not mean
+    the threshold public key is safe to use. Before using it, callers should
+    ensure that all participants have the recovery data, e.g., by collecting
+    acknowledgment signatures via `participant_recovery_ack_sign`.
```
real-or-random: "safe to use" may be unclear to the reader. I tried to expand on this:

```suggestion
    Since returning successfully does not imply that other participants deem the DKG session successful,
    returning successfully also does not imply that redundant copies of the recovery data exist.
    For example, it could be the case
    that other participants raised an exception instead, and this participant will be the
    only one that obtained the recovery data. In that case, if this participant's storage fails, 
    the only copy of the recovery data is lost
    As a result, this participant will not be able to convince any other participants to deem the DKG session successful, and it will not be possible to create a signature.
    
    To protect against this scenario, callers should ensure that all participants
    deem the DKG session successful (which also implies that they have a redundant copy of the recovery data)
    before using the threshold public key (e.g., before sending funds to it).  
    The recommended way of doing so is by collecting acknowledgment signatures via
    `participant_recovery_ack_sign`.
    Callers can alternatively employ some other means to
    ensure that they will always have access to the recovery data
    (which can be used to convince other participants that the DKG session was successful).
    For example, they could use a custom redundant way of backing up the recovery data.
```

What do you think? (the line lengths should be reformatted) Feel free to take only parts or modify further.

mllwchrry:
@real-or-random, I really like your explanation here. It removes the ambiguity of "safe to use" and makes the actual failure scenario concrete. Kept your wording and squashed everything into one commit.

real-or-random (approved):
ACK 125446149b4a51127f6d061a21cc30f5f84aa177

Issue #102, opened 2026-02-20
Clarify Lagrange identifier


jonasnick:
The BIP never explicitly says "the evaluation point for Lagrange interpolation in FROST is i+1, where i is the participant's 0-based index in hostpubkeys." A reader has to piece this together, and there's an easy off-by-one trap: line 507 says "their index" (which is i, 0-based), but the actual polynomial evaluation point is i+1.

We should explicitly states something like: "The participant with index i in hostpubkeys holds the secret share corresponding to polynomial evaluation point i+1. This evaluation point must be used as the participant's identifier when computing Lagrange coefficients in FROST signing sessions."

mllwchrry:
I agree that explicitly stating this is necessary. However, I found "corresponding to polynomial evaluation point i+1" a bit tricky to follow at first. Would it be clearer to say "computed by evaluating the polynomial at point i+1" instead? It seems more explicit about what's happening.

PR #111, opened 2026-04-08
Clarify Lagrange identifier

mllwchrry:
Addresses #102.


real-or-random:
I tried to phrase it more mathy (and thus hopefully clearer in #127). But I kept it in a nonessential footnote, because I also changed the text to say that this is designed only to work with BIP445. 

I think that's fine. People are their own if they use ChillDKG for other purposes than BIP445 then we give them a tiny warning at least. (To be honest, if they really repurpose the DKG, I hope they read the internals sections and the code...)

Issue #108, opened 2026-04-08
Minimal n


real-or-random:
In the BIP text, we say that `n >= 2` but the implementation doesn't make this restriction.

PR #112, opened 2026-04-08
Fix params validation to disallow n=1

DarkWindman:
Addresses #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,13 +140,15 @@ In summary, we aim for the following design goals:
 
  - **Standalone**: ChillDKG is fully specified, requiring no external secure channels or consensus mechanism.
  - **Conditional agreement**: If a ChillDKG session succeeds for one honest participant, this participant will be able to convince every other honest participant that the session has succeeded.
- - **No restriction on threshold**:  Like the FROST signing protocol, ChillDKG supports any threshold `t <= n`, including `t > n/2` (also called "dishonest majority").
+ - **No restriction on threshold**:  Like the FROST signing protocol, ChillDKG supports any threshold `t <= n`[^n-edge-cases], including `t > n/2` (also called "dishonest majority").
  - **Broad applicability**:  ChillDKG supports a wide range of scenarios, from those where the signing devices are owned and connected by a single individual to those where multiple owners manage the devices from distinct locations.
  - **Simple backups**: ChillDKG allows recovering the DKG output using the host secret key and common recovery data shared among all participants and the coordinator. This eliminates the need for session-specific backups, simplifying user experience.
  - **Untrusted coordinator**: Like FROST, ChillDKG uses a coordinator that relays messages between the participants. This simplifies the network topology, and the coordinator additionally reduces communication overhead by aggregating some of the messages. A faulty coordinator can force the DKG to fail but cannot negatively affect the security of the DKG.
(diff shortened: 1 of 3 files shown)


real-or-random:
I know I said that I think either way is fine. But now that I think about it again, I tend to think that allowing n=1 is better:
 - It's a nice corner case that is helpful for testing
 - Disallowing n=1 means more logic/more branches in the code. And we'll need additional test vectors that make sure that n=1 is rejected. 

What do you think?

DarkWindman:
Basically, I agree that n>1 restriction feels somewhat artificial, and it would not be a problem if we allow n=1. If users want to prevent 1-of-1, they can enforce that restriction at the application level. 
If we impose this restriction, it forces us to constantly mention it throughout the specification and maintain specific test vectors for it...

To sum-up, I think it would be better for us to support the 1-of-1 case. 
However, how should we mention this in the specification? I think we could update the requirement from `n >= 2` to `n >= 1` and add a footnote that 1-of-1 is a trivial case, which the protocol natively supports, and implementers are free to restrict it in their applications if desired (or just add a footnote).

real-or-random:
> However, how should we mention this in the specification? I think we could update the requirement from `n >= 2` to `n >= 1` and add a footnote that 1-of-1 is a trivial case, which the protocol natively supports, and implementers are free to restrict it in their applications if desired (or just add a footnote).

I thought about extending footnote 1 but I tend to think that the `t = n = 1` case is too trivial to even mention. And we currently say this:
> The case `t = 1` can be realized by letting one participant generate an ordinary [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) key pair and transmitting the key pair to every other participant, who can check its consistency and then simply use the ordinary [BIP 340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) signing algorithm.

Strictly speaking, this is correct also for the t = n = 1` case. ("every other participant" is vacuous then because there's no other participant).

DarkWindman:
> Strictly speaking, this is correct also for the t = n = 1` case. ("every other participant" is vacuous then because there's no other participant).

Yes, I agree that footnote 1 already adds clarification that the `t = n = 1` case is supported. However, we mention that in the introduction section, where we describe FROST. I think it would be better also if we add a footnote's 1 link in the 'Usage of ChillDKG' section, right after "There are `n >= 2` participants", to clarify that this restriction can be easily relaxed to `n >= 1`.

Overall, I do not think we even need to change footnote 1, since it already states that:

> While t = n and t = 1 are in principle supported, simpler alternatives are available in these cases. In the case of t = n, using a dedicated n-of-n multi-signature scheme such as MuSig2 [[BIP 327](https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki)] instead of FROST avoids the need for an interactive DKG.

real-or-random:
> right after "There are `n >= 2` participants", to clarify that this restriction can be easily relaxed to `n >= 1`.

Oh sure. My suggestion was simply to change this to `n >= 1` in the text without a footnote. Did this now in  4928fc2173f1c2bb22ebfe0fd0f34a5ec110c112.

DarkWindman:
Oh, okay! I agree. I think we can close this PR now, since the issue is addressed in commit `4928fc2`.

PR #113, opened 2026-04-14
test_vectors: Add group structure

DarkWindman:
This PR proposes a group structure for test vectors, allowing us to validate them across different threshold settings (previously, we only had `2-of-3`). 
In addition, we have divided the tests into two groups: those that require validation across multiple thresholds, and those that are independent of `t` and `n`. For the latter, we propose running them exclusively in a standard `2-of-3` configuration.

First, I would like to hear any suggestion on how to improve this construction. I believe this structure will be quite important for us.
Second, regarding the test messages: currently, I include the note "`t-of-n` group" in the description for tests running across different thresholds. Should we cover all of them this way, keep the current approach, or remove that information entirely?


python/gen_vector_utils/fixtures.py
```diff
@@ -0,0 +1,28 @@
...
+
+# Auxiliary randomness supplied by participant 0 during participant_step2.
+AUX_RAND_HEX = "005F5C3A69BB274F4559490AD754F1F5AFFABAED4C71AD5D8ACBAEFC2B491573"
+
+# (t, n, description) tuples defining the threshold configurations to test.
+THRESHOLD_CONFIGS = [
+    (2, 3, "Standard 2-of-3"),
+    (1, 3, "Minimum threshold 1-of-3"),
+    (3, 3, "Full-quorum 3-of-3"),
+    (3, 4, "Majority threshold 3-of-4"),
```
real-or-random: I doubt that the descriptions serve a good purpose.

real-or-random:
> In addition, we have divided the tests into two groups: those that require validation across multiple thresholds, and those that are independent of `t` and `n`. For the latter, we propose running them exclusively in a standard `2-of-3` configuration.

How do you make the judgment whether something is independent of `t` and `n` or not? Can you provide examples of either category?

I find it very difficult to navigate these big PRs. I suggest splitting the changes in multiple [small commits](https://web.archive.org/web/20111005024703/http://www.conifersystems.com/2008/11/05/the-benefits-of-small-commits/) that do one thing at a time.

> First, I would like to hear any suggestion on how to improve this construction. I believe this structure will be quite important for us.

This will depend on the answers above.

> Second, regarding the test messages: currently, I include the note "`t-of-n` group" in the description for tests running across different thresholds. Should we cover all of them this way, keep the current approach, or remove that information entirely?

I think it's good to have them, but just an auto-generated "t-of-n" suffices. No need to have nice descriptions like "standard".

DarkWindman:
> How do you make the judgment whether something is independent of t and n or not? Can you provide examples of either category?

For the independent set, I include tests that check for things like incorrect byte lengths, invalid values (deserialization/signature format checks), duplicate values, or infinity EC points. However, after double-checking, I realized I incorrectly categorized a few tests as independent. I will do another full pass to ensure everything is categorized correctly. If you have any other thoughts, I'd like to hear them!

> I suggest splitting the changes in multiple [small commits](https://web.archive.org/web/20111005024703/http://www.conifersystems.com/2008/11/05/the-benefits-of-small-commits/) that do one thing at a time.

Yes, I agree with you, that was my oversight! I will split the changes into smaller to make it easier to review.

real-or-random:
> I will do another full pass to ensure everything is categorized correctly. If you have any other thoughts, I'd like to hear them!

Makes sense. 

If it turns out that there are not many "independent" tests, then it may be simpler to run all of them for all values of (n,t). (Not implying that this is necessarily the case, I think I need to look at it again after your updates.)

DarkWindman:
I reviewed the test vectors again yesterday, and then discussed them with Maria. We realized our previous approach to classifying tests by threshold (`t, n`) dependencies is not correct. We concluded that we really only need to run tests across different threshold configurations if the threshold actually affects the test logic. In practice, this mostly applies to the valid test cases.
To be more precise: tests verifying the validity of a participant's Step 1 should absolutely be tested across different thresholds. However, error tests, such as a participant sending an invalid signature or an encrypted secshare, do not need this. Whether it is 2-of-3 or 1-of-3, the validation logic is exactly the same: the array contains an invalid value, and the function correctly returns an error.

Unless we are missing an edge case, this means we can significantly simplify the group structure. @real-or-random, what do you think? I'm currently looking for any new specific error vectors that would behave differently depending on the threshold, but if there aren't any, I'll proceed with the simplified structure I started yesterday.

real-or-random:
> I'm currently looking for any new specific error vectors that would behave differently depending on the threshold,

I think we need to distinguish between
1. what the test logic does
2. what our reference implementation does
3. what other, potentially buggy, implementations could do

3 is the main purpose of the test vectors. And here, all functions that "know" t and n could potentially branch on them for some reason, even if this is wrong to do. 

How bad would it be to keep it as simple as possible and run all vectors for all threshold, even if means that some tests are redundant? This would keep the test logic simple, and make it easier for other implementations to reimplement it without missing some cases accidentally.

DarkWindman:
> How bad would it be to keep it as simple as possible and run all vectors for all threshold, even if means that some tests are redundant? 

I think we were initially just worried about redundant test setups causing duplication. In reality, it wouldn't be a catastrophe if we did that. It would only increase the size of the JSON vector files. We can simply run all vectors across all thresholds and keep the implementation as simple as possible.

real-or-random:
> We can simply run all vectors across all thresholds and keep the implementation as simple as possible.

Let's do this first. I don't expect that the JSON files will be super large, or that the running time will be too high. But if that ends up being the case nevertheless, then we can still see how we can improve.

DarkWindman:
Now, all participant and coordinator test vectors are generated across different thresholds. For better JSON readability, we indicate the threshold at the beginning of each group.

siv2r:
I like the current design. Each vector file has a `testGroups` array, each group represents one `(t, n)` config, and all the cases for that config live inside it.

<details>
<summary>JSON (current design)</summary>

```json
{
  "testGroups": [
    {
      "t": 2, "n": 3,
      "hostseckeys": ["...", "...", "..."],
      "hostpubkeys": ["...", "...", "..."],
      "tests": [
        { "comment": "valid participant step1",
          "input": { "hostseckey": "...", "randomness": "..." },
          "expected": { "pmsg1": "..." } },
        { "comment": "hostseckey wrong length",
          "input": { "hostseckey": "ab..." },
          "error": { "type": "HostSeckeyError" } },
        { "comment": "t=0",
          "input": { "t": 0 },
          "error": { "type": "ThresholdOrCountError" } }
      ]
    },
    { "t": 1, "n": 3, "hostseckeys": [...], "tests": [ /* same cases, different bytes */ ] },
    { "t": 3, "n": 3, "hostseckeys": [...], "tests": [ /* same cases, different bytes */ ] }
  ]
}
```

</details>

I did consider an alternative design: instead of grouping by config, group by test case. Each case carries its per-config inputs and expected outputs side by side:

<details>
<summary>JSON (case-grouped alternative)</summary>

```json
{
  "configs": [
    { "t": 2, "n": 3 },
    { "t": 1, "n": 3 },
    { "t": 3, "n": 3 }
  ],
  "tests": [
    { "comment": "valid participant step1",
      "per_config": {
        "(2,3)": { "input": {...}, "expected": {...} },
        "(1,3)": { "input": {...}, "expected": {...} },
        "(3,3)": { "input": {...}, "expected": {...} }
      } },
    { "comment": "hostseckey wrong length",
      "input": {...},
      "error": {...} }
  ]
}
```

</details>

The advantage I can see is review-side. You can read one case and immediately see how its inputs and outputs differ across configs without flipping between groups. Easier to audit what a test case is for.

But it complicates the test harness. The current design lets the harness loop `testGroups → tests` and run each case directly. The case-grouped alternative makes it look up the matching config and fixtures per case before running it, which adds indirection without much gain on the implementer side. So I think the current design is better.

I also checked how other crypto protocols handle multi-config test vectors, and none of them vary `(t, n)` across their vectors:

- [Wycheproof](https://github.com/C2SP/wycheproof/tree/main/testvectors_v1): lots of parameter variation across primitives, but no threshold or multi-party protocols
- [RFC 9591 (FROST)](https://www.rfc-editor.org/rfc/rfc9591.html#name-test-vectors): single `(t, n)` per ciphersuite, no parameterization.
- [NIST ACVP](https://pages.nist.gov/ACVP/): `testGroups` idiom for parameter variation, similar to our current design.
- [HPKE](https://www.rfc-editor.org/rfc/rfc9180.html#name-test-vectors) / [MLS](https://github.com/mlswg/mls-implementations/tree/main/test-vectors): one secret key & config used for all test cases.

So ChillDKG would be the first-mover here. The current design is a reasonable extension of the Wycheproof `testGroups` style for our use case. 

----

Side note 1: right now we only vary `t` (n stays at 3 across all three configs). I think we could vary `n` too. The input is just hostseckeys, one per participant, and they're independent, so scaling `n` is just adding more keys to the list. Worth considering for better coverage of `n`.

Side note 2 (nit): In this RFC [test vectors](https://www.rfc-editor.org/rfc/rfc9807.html#name-test-vectors). They follow:

<details>
<summary>JSON </summary>

```json
{
  "testGroup": {
    "config": ["t":2, "n":3],
    "input_values": {
      "param1": "...",
      "param2": "...",
    },
    "intermediate_values": {
      "param1": "...",
      "param2": "...",
    },
    "ouput_values": {
      "param1": "...",
      "param2": "...",
    }
  }
}
```

</details>

We could adopt the `intermediate value` field to our `_step2.json` vectors as it makes the file more readable (but adds a extra nesting wrt to the testing harness). Tbh, this just gives a minor improvement as the `description` field in our file calls out the intermediate functions nicely.

python/gen_vector_utils/participant.py
```diff
@@ -15,43 +15,38 @@
...
+    "For each error test case:",
+    "  Call participant_step1(hostseckey, params, random).",
+    "  Verify it raises an exception matching expected_error.",
+]
+
+
+def generate_participant_step1_group(t, n):
     valid_cases = []
     error_cases = []
     tc_id = 0
```
siv2r: `tc_id` needs to be unique across all test cases in a file so each one can be identified without ambiguity. Right now, `tc_id` resets to `0` at the start of each `_group` function, so duplicate IDs end up across groups in the same vector file.

We can change each `_group` function to take a `tc_id_init` parameter as the starting value and increment from there, so IDs stay unique across groups.

siv2r: nit: we could introduce `tg_id` to identify each group. I've seen this pattern being used in test cases of other protocols. Not strictly neccessary, but helps with debugging.

siv2r:
Right now this PR introduces a `testGroups` array inside each vector file, with one group per `(t, n)` config (`2-of-3`, `1-of-3`, `3-of-3`). I want to raise an alternative before we settle on this.

What if we keep one file per config instead? So instead of bundling every threshold configs into `participant_step1_vectors.json` under `testGroups`, we'd have `2-of-3-participant-step1.json`, `1-of-3-participant-step1.json`, and so on. Each file goes back to the flat `valid_test_cases` / `error_test_cases` shape we already had, and we drop the `testGroups` wrapper entirely.

A few things I like about this. We don't have to complicate the existing files with the extra nesting. If an implementation only cares about testing `2-of-3-participant-step1.json`, they can just run that one and ignore the `1-of-3` file. And it doesn't really complicate the harness either, since every file has the same structure, so we just loop over the list of files instead of looping over groups inside a file.

The one downside I see is that we'd generate a lot more vector files. But honestly that feels fine to me. Most protocols I've looked at ship a whole bunch of vector files anyway, so it's a pretty normal pattern.

What do you all think of this design? I don't have a strong preference, just wanted to raise it before we settle on `testGroups`.

python/gen_vector_utils/participant.py
siv2r: The current `threshold` field servers the neccessity of `tg_id`. It's values (`2of3`, `1of3`) are both human readable, and unique across test groups. So, we don't need `tg_id`, or we could rename `threshold` to `tg_id`.

mllwchrry:
@siv2r, thanks a lot for the detailed comments!

Among the two designs you proposed, we actually lean toward a file-per-config approach. It looks the cleanest compared to keeping everything in one big file, and could be easily done with the current code.

> I think we could vary n too.

Good catch, agree. Will do this once we agree on the design.

> We could adopt the intermediate value field to our _step2.json vectors as it makes the file more readable (but adds a extra nesting wrt to the testing harness). Tbh, this just gives a minor improvement as the description field in our file calls out the intermediate functions nicely.

The idea is good overall. For _step2.json, the most natural candidate would be `eq_input`. It's accessible from the returned state without modifying the reference implementation. However, it's already implicitly present in the finalize vectors within `recovery_data`. So adding it explicitly would just make it more accessible, not new information. Are there other specific fields you had in mind? So that we can decide whether it is worth it.

@real-or-random, what do you think about all this? Would like to know your opinion before we move to modifying the code again.

siv2r:
> we actually lean toward a file-per-config approach.

Yes, I think the multi-file approach is a good fit, especially for the `_step2` vectors, since they effectively exercise the entire protocol.

For the signing BIP, I decided to stick with `test_groups` rather than multiple files because those vectors are just testing a single function.

> Are there other specific fields you had in mind? So that we can decide whether it is worth it.

I don't have any concrete suggestions at the moment. Out of the various patterns I've seen, this one seemed potentially useful for DKG, so I thought I'd mention it. The benefit is fairly small and it does add some extra nesting, so leaving it out is fine.

real-or-random:
First of all: I'm really sorry for my (way too) long silence here.

I very much like the current state of the PR. Here are some questions/minor things. None of these are blockers, and I don't have strong opinions on any of these choices.

#### Varying n

Yes, this is a good idea.

#### Redundancy
  
* I don't think the "intermediate values" thing adds much. It makes the JSON files a tiny bit more readable, but at the cost of adding redundancy in the data. And these files are primarily meant to be read by machines, and avoiding redundancy is the right thing to do when optimizing for machines.
* Relatedly, I think we could drop the `threshold` field because it's redundant. We have `t` and we have a list of hostpubkeys.

#### File Structure

I tend prefer the file structure as in the current state of the PR (one JSON file per tested function).

I think when we should optimize primarily for convenience of use (by implementations), i.e., we should make it easy to write tests that exercise the vectors. And how would you do this? The actual test code will depend very much on the tested function (but not really much on the threshold). So the natural structure of tests is to have one test for each tested function. Then each test needs to read exactly one JSON file and run the checks. This is what `tests.py` in the current state of the PR does, and I think it's a nice and natural structure.

Making more files means that tests need extra complexity for discovering test files, e.g., globs such as `participant_step1_*.json` (or loops over t and n) to discover the test files. This is easy to get wrong. If your glob matches nothing, you don't read any JSON files and "all" tests will pass.

The opposite, i.e., having everything in a single file would also be fine. But again, this could make it easier for the test implementer to miss or overlook some tests. 

#### Grouping 

@siv2r has nicely explained why the current form is better for the test implementer, and as I said, I think this is what we should optimize for. So yes, let's keep the current form.

#### Wycheproof Conventions
I think it's a good thing to follow Wycheproof conventions. (If anything, for the lack of another established convention.) But if we follow them, then we should do it consistently.  
 - `testGroups` vs. `test_groups`: We usually prefer snake_case everywhere else, but camelCase appears to be the  Wycheproof convention, e.g., `testGroups` and `tcId`
 - They don't have a `tgId` so I don't think we should have one.
 - They have JSON schemas for some of their vectors, see https://github.com/C2SP/wycheproof/tree/main/schemas My feeling is that this is borderline overkill, but if you want to go down that road, I wouldn't be opposed to it.

mllwchrry:
All review comments have been addressed.

One open question: `testGroups` / `tcId` vs `test_groups` / `tc_id`. I went with snake notation throughout the json to stay consistent with the rest of the reference implementation. Switching to camelCase for just these two fields felt inconsistent. However, if you prefer to follow Wycheproof conventions strictly, I'm happy to migrate all json field names to camelCase.

Once we agree on the current version, I'll resolve the conflicts and the PR will be ready to merge.

mllwchrry:
All review comments have been addressed (except for the notation question, see above) and conflicts resolved.  @real-or-random, requesting another round of review.

real-or-random:
> However, if you prefer to follow Wycheproof conventions strictly, I'm happy to migrate all json field names to camelCase.

This is what I tried to suggest above in my previous comment, because otherwise it does not make much sense to use the  names from Wycheproof. 

edit: I'll do another round of review tomorrow in my morning.

mllwchrry:
> This is what I tried to suggest above in my previous comment, because otherwise it does not make much sense to use the names from Wycheproof.

Fixed in 51377f3.

python/gen_vector_utils/coordinator.py
```diff
@@ -153,15 +155,47 @@ def generate_coordinator_step1_vectors():
...
         "valid_test_cases": valid_cases,
         "error_test_cases": error_cases,
     }
 
 
-def generate_coordinator_finalize_vectors():
+def generate_coordinator_step1_vectors():
+    groups = []
+    group = generate_coordinator_step1_group()
+    tc_id = len(group["valid_test_cases"]) + len(group["error_test_cases"])
```
real-or-random: Shouldn't this variable better be called `total_tests`?

(The same is true in similar functions.)

real-or-random (approved):
Looks great.

One more suggestion: The tests in tests.py could also assert that the tcId is correctly set/incremented. This is not essential for testing implementation but it ensures that our generation scripts do the right thing.

ACK otherwise

mllwchrry:
@real-or-random, fixed your recent comments, thanks.

real-or-random (approved):
utACK 9d3ccf30c60920152a810eceff20dd035ea39704

Issue #114, opened 2026-04-29
ValueError in public API functions is undocumented and uninformative


mllwchrry:
Currently, the code raises `ValueError` in the new public function `participant_recovery_acks_verify`, similar to how we do it for length checks in other public functions (`coordinator_step1` and `coordinator_finalize`). However, this exception is never documented in the spec for any of these functions and seems to be non-informative.
Should we open a separate PR to address this? We could either document `ValueError` in the Raises sections of all three functions, or create more specific exceptions.

_Originally posted by @mllwchrry in https://github.com/BlockstreamResearch/bip-frost-dkg/issues/106#issuecomment-4178865450_

real-or-random:
> Currently, the code raises `ValueError` in the new public function `participant_recovery_acks_verify`

Strictly speaking, the code doesn't "currently" raise, because #106 is not yet merged.

Let's just discuss this in #106.  ;)

real-or-random:
Ah sorry, nevermind. You're saying it's a general problem in multiple functions, including functions we currently have.

real-or-random:
Do you think it would be clearer to raise an `IndexError`? In either case, do you think it's better to use a built-in exception or create a more specific exception?

mllwchrry:
`IndexError` feels more like an array access issue. As for me, `ValueError` is the only remotely fitting built-in exception. However, I lean toward a custom exception, because it would be more informative and consistent with, e.g., existing `HostSeckeyError` and `RandomnessError`.
I have 3 name candidates, but would be happy to hear other options: `InputLengthError`, `ParticipantListLengthError`, `WrongNumberOfInputsError`.

real-or-random:
> `IndexError` feels more like an array access issue. As for me, `ValueError` is the only remotely fitting built-in exception. However, I lean toward a custom exception, because it would be more informative and consistent with, e.g., existing `HostSeckeyError` and `RandomnessError`. 

I tend to agree with making it consistent with `HostSeckeyError` and `RandomnessError`, though there's another dimension here. These errors are raised if the user passes obviously inputs which have the wrong length. This means they're "structurally" invalid. (Wrong type would be another example of this.) The way I think about these is that they're not really part of the ChillDKG protocol, and dealing with them is not the primary goal of our reference implementation. Strictly speaking fact, we wouldn't need to check inputs at all because the docs tell you what to pass, and we don't need to make guarantees for the case of wrong inputs.

[...]

PR #118, opened 2026-06-10
Refine input validation and error handling

DarkWindman:
Addresses the suggestion from #114.

- Differentiate structural errors (`ValueError`) from protocol-level errors (`HostSeckeyError`, `RandomnessError`). 
- Update test vectors covering invalid hostseckeys for all functions taking one, plus all-zero random for `participant_step1`.
- Add `participant_step2` check that hostseckey matches the one used in `participant_step1`.


README.md
```diff
@@ -895,14 +892,13 @@ Perform a participant's first step of a ChillDKG session.
...
-  key is invalid, or if the key does not match any entry of
-  `hostpubkeys`.
+- `HostSeckeyError` - If the key is invalid, or if the key does not match any
+  entry of `hostpubkeys`.
 - `InvalidHostPubkeyError` - If `hostpubkeys` contains an invalid public key.
 - `DuplicateHostPubkeyError` - If `hostpubkeys` contains duplicates.
 - `ThresholdOrCountError` - If `1 <= t <= len(hostpubkeys) <= 2**32 - 1` does
   not hold.
-- `RandomnessError` - If the length of `random` is not 32 bytes.
+- `RandomnessError` - If `random` is all zeroes.
```
real-or-random: It would be nice to add a sentence of explanation/rationale here.

python/gen_vector_utils/participant.py
```diff
@@ -77,7 +78,41 @@ def generate_participant_step1_group(t, n, tc_id_init=0):
             "comment": "length of host secret key is not 32 bytes",
         }
     )
-    # --- Error test case 1: Invalid threshold ---
+    # --- Error Test Case 1: zeroed hostseckey ---
```
real-or-random: nit: s/zeroed/zero

python/gen_vector_utils/participant.py
```diff
@@ -94,7 +129,7 @@ def generate_participant_step1_group(t, n, tc_id_init=0):
             "comment": "invalid threshold value",
         }
     )
-    # --- Error test case 2: hostpubkeys list contains an invalid value ---
+    # --- Error test case 4: hostpubkeys list contains an invalid value ---
```
real-or-random: I suggest to drop the numbers in the comments entirely. The seem to hard to maintain and they pollute diffs

python/chilldkg_ref/chilldkg.py
mllwchrry:
```suggestion
    """"Raised if the provided randomness is all zeroes."""
```

python/gen_vector_utils/participant.py
siv2r: Agreed, and the current behavior is a bit misleading since `tc_id` changes across groups.

On a related note, we could remove `tc_id += 1` from the generator files entirely and have a single function assign them ([like this](https://github.com/siv2r/bip-frost-signing/blob/master/python/generators/common.py#L292-L299)). Simplifies the id assignment.

DarkWindman:
Addressed review comments and added an extra test vector for the `aux_rand` length check in `participant_step2`.

python/chilldkg_ref/chilldkg.py
```diff
@@ -524,16 +526,16 @@ def participant_step1(
...
-            key is invalid, or if the key does not match any entry of
-            `hostpubkeys`.
+        HostSeckeyError: If the key is invalid, or if the key does not match any
+            entry of `hostpubkeys`.
         InvalidHostPubkeyError: If `hostpubkeys` contains an invalid public key.
         DuplicateHostPubkeyError: If `hostpubkeys` contains duplicates.
         ThresholdOrCountError: If `1 <= t <= len(hostpubkeys) <= 2**32 - 1` does
             not hold.
-        RandomnessError: If the length of `random` is not 32 bytes.
+        RandomnessError: If `random` is all zeroes. This check guards against the
```
real-or-random:
```suggestion
        RandomnessError: If `random` is all zeroes (i.e., b"\x00" * 32). This check guards against the
```

This makes it a tiny bit clearer. "All zeroes" could also mean `b"0" * 32` or whatever (though this would be weird in this context...). Sorry, I suggested this terminology. It's not uncommon, but also not super precise.


(You may need to reformat the paragraph.)

python/gen_vector_utils/participant.py
DarkWindman: > On a related note, we could remove tc_id += 1 from the generator files entirely and have a single function assign them ([like this](https://github.com/siv2r/bip-frost-signing/blob/master/python/generators/common.py#L292-L299)). Simplifies the id assignment.

Good suggestion, agreed. However, I think this should be addressed in a separate PR.

mllwchrry:
ACK 90c94f8

python/chilldkg_ref/chilldkg.py
```diff
@@ -524,16 +526,16 @@ def participant_step1(
         bytes: The first message to be sent to the coordinator.
 
     Raises:
-        HostSeckeyError: If the length of `hostseckey` is not 32 bytes, if the
-            key is invalid, or if the key does not match any entry of
-            `hostpubkeys`.
+        HostSeckeyError: If the key is invalid, or if the key does not match any
```
real-or-random: nit:
If we anyway touch these lines, we could say "host secret key" instead of key.

python/chilldkg_ref/chilldkg.py
DarkWindman: Oh yes, agreed! This is more precise.

real-or-random:
Your last commit screwed up README.md somehow but otherwise ACK

python/chilldkg_ref/chilldkg.py
```diff
@@ -558,15 +562,15 @@ def participant_step1(
...
+    )
 
     state1 = ParticipantState1(params, idx, enc_state)
     pmsg1 = enc_pmsg
     return state1, pmsg1
 
 
 class RandomnessError(ValueError):
-    """Raised if the length of the provided randomness is not 32 bytes."""
+    """Raised if the provided randomness is all zeroes."""
```
real-or-random: Then this should be changed accordingly:
```suggestion
    """Raised if the provided randomness is all zeroes (i.e., b"\\x00" * 32)."""
```

Sorry, I promise that this is my last nit in this PR. :D

real-or-random (commented):
ACK 530a4abf84176142df7934d7952fbaceed4291c8

Issue #117, opened 2026-05-10
Test vector error message mismatches comment


aakselrod:
Perhaps it's better to split [this test case](https://github.com/BlockstreamResearch/bip-frost-dkg/blob/ebff051bb6c6d9787cb0252e63595bf920276754/python/vectors/participant_step2_vectors.json#L49) into two? If the implementation actually [parses the pubnonce](https://codeberg.org/aakselrod/btcd/src/commit/6298bea9a0695e13ce4aaebd812c6272842ee722/btcec/schnorr/frost/encpedpop/encpedpop.go#L283-L286) (please forgive the current state of the linked implementation) into a public key object while parsing the message, the error message will match the comment. I propose a split like:

- First one with only an invalid public nonce, and the rest of the message correct
- Second with all valid public nonces but missing/undersized encrypted secret shares

mllwchrry:
Good catch, thank you! I agree with the proposed fix: a clean test for missing encrypted secret shares and a separate invalid pubnonce test. The latter is already planned in an upcoming PR. We'll add both after #113 is merged to avoid file conflicts.

mllwchrry:
Fixed in #119. @aakselrod, does this address your concern? Feel free to close the issue if so.

real-or-random:
Let's assume this has been fixed in #119.

 @aakselrod Please tell us if you have further comments.

PR #119, opened 2026-06-11
python: Add missing test vectors for edge cases and fix comment typo

mllwchrry:
## Summary

- Add `t` > `n` error case to `participant_step1` vectors.
- Split `participant_step2` error case into two cases: invalid pubnonce value and missing encrypted secret share as proposed in #117.
- Add duplicate pubnonces check to `participant_step2` vectors.
- Add too-long length tests complementing the existing too-short tests (forces `!=` rather than `<`/`>` length checks): too long `cmsg2` in `participant_finalize`, too many `pmsgs2` in `coordinator_finalize`.
- Fix `coms_to_secret` to `coms_to_secrets` typo in comments.


python/vectors/coordinator_finalize_vectors.json
```diff
@@ -80,6 +80,19 @@
...
+                    "pmsg2Indices": [
+                        0,
+                        1,
+                        2,
+                        0
+                    ],
+                    "expectedError": {
+                        "type": "ValueError"
+                    },
+                    "comment": "4 pmsgs2 provided instead of 3"
```
real-or-random: I think this can be a hard to parse for a human being, because `pmsgs2` is not really a plural (though it's the correct parameter name) and it's actually an array. 

Here's a suggestion that roughly follows the other comments: 
`Invalid pmsgs2: more entries than participants`

python/gen_vector_utils/participant.py
```diff
@@ -249,7 +266,7 @@ def generate_participant_step2_group(t, n, tc_id_init=0):
     # --- Error Test Case 0: pubnonces list in cmsg1 has an invalid value at index 0 ---
     tc_id += 1
     invalid_cmsg1_parsed = copy.deepcopy(cmsg1_parsed)
-    invalid_cmsg1_parsed.enc_cmsg.pubnonces[0] = b"\xeb" * 32  # random pubnonce
+    invalid_cmsg1_parsed.enc_cmsg.pubnonces[0] = b"\xeb" * 33  # invalid pubnonce
```
real-or-random, line 355: nice catch...

real-or-random (commented):
Beyond this PR, I wonder if we should somehow treat these `ValueError` tests differently. My thinking is that we should certainly have them. But it can be a bit weird for the implementer. Depending on their programming language, some of them won't matter. Maybe we should add a comment somewhere. And/or group them together, so they can be easily skipped as a whole.

python/vectors/coordinator_finalize_vectors.json
mllwchrry: Fixed in 020b19e

real-or-random:
This needs rebase now after #118  :/

mllwchrry:
> This needs rebase now after #118 :/

We were expecting this. Rebased.

python/gen_vector_utils/coordinator.py
```diff
@@ -249,7 +249,22 @@ def generate_coordinator_finalize_group(t, n, tc_id_init=0):
             "tcId": tc_id,
             "pmsg2Indices": list(range(len(pmsgs2) - 1)),  # [0, ..., n - 2]
             "expectedError": error_case,
-            "comment": f"only {len(pmsgs2) - 1} pmsgs2 provided instead of {len(pmsgs2)}",
+            "comment": "invalid pmsgs2: fewer entries than participants",
+        }
+    )
+
+    # --- Error test case: too many pmsgs2 (n+1 instead of n) ---
```
DarkWindman: I think we should rename it for consistency.

```suggestion
    # --- Error test case: invalid pmsgs2 (n+1 entries instead of n) ---
```

python/gen_vector_utils/participant.py
```diff
@@ -347,7 +364,39 @@ def generate_participant_step2_group(t, n, tc_id_init=0):
...
+            "tcId": tc_id,
+            "cmsg1": bytes_to_hex(invalid_cmsg1),
+            "expectedError": error,
+            "comment": "invalid cmsg1: missing encrypted secret shares",
+        }
+    )
+    # --- Error test case: pubnonces list in cmsg1 has duplicate values ---
+    tc_id += 1
+    invalid_cmsg1_parsed = copy.deepcopy(cmsg1_parsed)
+    invalid_cmsg1_parsed.enc_cmsg.pubnonces[1] = invalid_cmsg1_parsed.enc_cmsg.pubnonces[0]
```
DarkWindman:
```suggestion
    invalid_cmsg1_parsed.enc_cmsg.pubnonces[1] = (
         invalid_cmsg1_parsed.enc_cmsg.pubnonces[0]
     )
```

python/gen_vector_utils/participant.py
real-or-random: Hm, now that I see this again, I think the original purpose of the test vector was something else.

If I remember correctly, we had 32-byte pubnonces at some point in the past. So `b"\xeb" * 32` was simply the wrong value, whereas `b"\xeb" * 33` is really "invalid" already because it doesn't start with `0x02` or `0x03`. 

It's probably enough to test wrong values, but my thinking is that we could just two separate tests for the two cases.

mllwchrry: Good point. The original `b"\xeb" * 32` was changed to `b"\xeb" * 33` to fix the accidental length change it caused (the 32-byte value shortened the message, so the error was triggered by a length check rather than the pubnonce value itself). The duplicate pubnonces test already covers the "valid point but wrong value" case, so together they cover both scenarios you mentioned. Or did I misunderstand you?

real-or-random: (I think you understand me correctly.)

>  The duplicate pubnonces test already covers the "valid point but wrong value" case, so together they cover both scenarios you mentioned. 

I see! But then, given that we have this "duplicate" vector, it's conceivable that an implementer thinks there should be a duplicate check and implements this directly. In that case, it will be nice to have another test vector that simply has a hardcoded wrong point. (With that in mind, we could also remove the "duplicate" vector and only have a "wrong" vector; not sure what's better.) 

And for wording of the comments: I see that "random" was not good, but I think also "invalid" could make people think "invalid point". "Wrong" is not much better. Maybe "hardcoded" or "arbitrary hardcoded"? 


*edit: Sorry, now I am confused. Let me say in new words: I think there should be at least one case with an invalid point, and one case with a "wrong" value (preferably not the "duplicate" one). 

And Are you saying that `b"\xeb" * 32` is an invalid point? (I didn't check.) Then let's call this "invalid point" in the comments, and call the other one "hardcoded" or "arbitrary hardcoded".*

mllwchrry: Agree. A new test for valid-but-wrong EC point is needed, and I'll add it. As for the duplicate test, I think it's better to keep it, since it raises `UnknownFaultyParticipantOrCoordinatorError`, while a valid-but-wrong EC point raises `FaultyCoordinatorError`. So, it exercises a different code path.

Regarding the wording, I was also considering "mismatched", but "arbitrary" seems the cleanest. Would "arbitrary" alone suffice, or do you prefer "arbitrary hardcoded"?

real-or-random: > I was also considering "mismatched", but "arbitrary" seems the cleanest. Would "arbitrary" alone suffice, or do you prefer "arbitrary hardcoded"?

Any one of these three variants sounds good to me.

mllwchrry: Added a test with an arbitrary valid-but-wrong EC point in d868c7a. Also I extracted it as a named constant `ARBITRARY_POINT_X`, and updated all places that said "invalid" but meant "arbitrary".

real-or-random (approved):
ACK d868c7a80cb9e4adf4b32e88ca0e3319224bbc22

python/gen_vector_utils/participant.py
```diff
@@ -347,12 +366,10 @@ def generate_participant_step2_group(t, n, tc_id_init=0):
...
         }
     )
-    # --- Error test case: coms_to_secret list in cmsg1 has an invalid value at index 0 ---
+    # --- Error test case: pubnonces list in cmsg1 has an arbitrary value at index 0 ---
     tc_id += 1
     invalid_cmsg1_parsed = copy.deepcopy(cmsg1_parsed)
-    invalid_cmsg1_parsed.enc_cmsg.simpl_cmsg.coms_to_secrets[0] = GE.lift_x(
-        0x60C301C1EEC41AD16BF53F55F97B7B6EB842D9E2B8139712BA54695FF7116073
-    )  # random GE
+    invalid_cmsg1_parsed.enc_cmsg.pubnonces[0] = GE.lift_x(ARBITRARY_POINT_X).to_bytes_compressed()
```
DarkWindman:
```suggestion
    invalid_cmsg1_parsed.enc_cmsg.pubnonces[0] = GE.lift_x(
        ARBITRARY_POINT_X
    ).to_bytes_compressed()
```

python/gen_vector_utils/participant.py
```diff
@@ -363,10 +380,61 @@ def generate_participant_step2_group(t, n, tc_id_init=0):
...
+            "tcId": tc_id,
+            "cmsg1": bytes_to_hex(invalid_cmsg1),
+            "expectedError": error,
+            "comment": "invalid cmsg1: missing encrypted secret shares",
+        }
+    )
+    # --- Error test case: coms_to_secrets list in cmsg1 has an arbitrary value at index 0 ---
+    tc_id += 1
+    invalid_cmsg1_parsed = copy.deepcopy(cmsg1_parsed)
+    invalid_cmsg1_parsed.enc_cmsg.simpl_cmsg.coms_to_secrets[0] = GE.lift_x(ARBITRARY_POINT_X)
```
DarkWindman:
```suggestion
    invalid_cmsg1_parsed.enc_cmsg.simpl_cmsg.coms_to_secrets[0] = GE.lift_x(
        ARBITRARY_POINT_X
    )
```

python/gen_vector_utils/participant.py
```diff
@@ -755,9 +836,7 @@ def generate_participant_investigate_group(t, n, tc_id_init=0):
             cinv_msgs[0], len(params.hostpubkeys)
         )
         invalid_cinv_msg0_parsed = copy.deepcopy(cinv_msg_parsed)
-        invalid_cinv_msg0_parsed.enc_cinv.partial_pubshares[1] = GE.lift_x(
-            0x60C301C1EEC41AD16BF53F55F97B7B6EB842D9E2B8139712BA54695FF7116073
-        )  # random GE
+        invalid_cinv_msg0_parsed.enc_cinv.partial_pubshares[1] = GE.lift_x(ARBITRARY_POINT_X)
```
DarkWindman:
```suggestion
        invalid_cinv_msg0_parsed.enc_cinv.partial_pubshares[1] = GE.lift_x(
            ARBITRARY_POINT_X
        )
```

DarkWindman (commented):
Some not critical nits.

DarkWindman:
ACK https://github.com/BlockstreamResearch/bip-frost-dkg/commit/62b05759ea06675c7886973c3a4c8af9b0805f1e. I'll fix the failed tests in the next PR.

PR #110, opened 2026-04-08
Add on-wire message sizes to ChillDKG sequence diagram

mllwchrry:
Adds on-wire byte sizes to each arrow in the ChillDKG sequence diagram, resolving the `TODO` in `README.md`.

PR #123, opened 2026-06-18
python: Fix wrong-length tests in test_recovery_acknowledgment

DarkWindman:
Merging #106 (recovery acknowledgment) after #118, left some logic in test_recovery_acknowledgment outdated.

### DIFF
--- a/python/chilldkg_ref/chilldkg.py
+++ b/python/chilldkg_ref/chilldkg.py
@@ -141,9 +141,7 @@ def recovery_ack_message(x: bytes, idx: int) -> bytes:
     return prefix + idx.to_bytes(4, "big") + x
 
 
-def recovery_ack_sign(
-    hostseckey: bytes, idx: int, x: bytes, aux_rand: bytes
-) -> bytes:
+def recovery_ack_sign(hostseckey: bytes, idx: int, x: bytes, aux_rand: bytes) -> bytes:
     msg = recovery_ack_message(x, idx)
     return schnorr_sign(msg, hostseckey, aux_rand=aux_rand)
 
@@ -1058,7 +1056,10 @@ class RecoveryDataError(ValueError):
 
 
 def participant_recovery_ack_sign(
-    hostseckey: bytes, recovery_data: RecoveryData, params: SessionParams, aux_rand: bytes
+    hostseckey: bytes,
+    recovery_data: RecoveryData,
+    params: SessionParams,
+    aux_rand: bytes,
 ) -> bytes:
     """Sign recovery data to create a recovery acknowledgment.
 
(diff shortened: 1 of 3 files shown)


python/chilldkg_ref/chilldkg.py
```diff
@@ -1163,7 +1163,7 @@ def participant_recovery_acks_verify(
 
     for i, sig in enumerate(ack_sigs):
         if len(sig) != 64:
```
mllwchrry: Shouldn't this remain `InvalidRecoveryAckError(i)`? Changing to `ValueError` is consistent with the length-check policy from #118, but here we know exactly which participant is faulty (unlike the wrong-count check above). Every other per-participant malformed message in the codebase raises a faulty participant error rather than `ValueError`. However, I'm unsure about it, @real-or-random, what do you think?

python/chilldkg_ref/chilldkg.py
mllwchrry, line 1166: Pre-existing nit: this check is not covered by a test. We could add it in this PR.

DarkWindman: Good catch!

python/chilldkg_ref/chilldkg.py
DarkWindman: Hmm, after giving it some thought, I agree this fits the existing per-participant malformed-message pattern (like wrong-length pmsg in coordinator_step1). The per-sig length is per-participant, so we should attribute the error to them. In contrast, the wrong-count check above is genuinely caller-side.

Fixed in f4e065d1e653de99cc81c8229741cb2bf8d94f40, but let's leave the final word to @real-or-random.

real-or-random: The way I see the discussion in #118 is that it's about "real" inputs from the caller, not about received network messages. What we do in the rest of the codebase for network message is to try to parse them. If this fails, we catch the corresponding `MsgParseException` and raise a different API-level exception. And yes, `InvalidRecoveryAckError` would be appropriate here then.

What I notice now is that we didn't introduce a wrapper class `RecoveryAckMsg` for recovery ack messages, with parsing functions. I admit it's a bit pointless since it just wraps a single thing. But we did this also in `ParticipantMsg2` because we had a class for every other message. So I think this should be done also here for consistency (in this PR or in another PR). 

Beyond this PR:
But this raises a general question: what are these things that we call messages and what is the caller supposed to do with them? At the moment, these are variable-length byte strings at the code level, even though their length is statically determined by n and t (see #110). This is not very useful for the caller because the caller doesn't know how many bytes to except over the network. In each `...Msg` class, we should give the caller give a method that take (n, t) and outputs the length. (And suddenly, these "singleton" wrapper classes mentioned above make more sense. ;))

DarkWindman: > What I notice now is that we didn't introduce a wrapper class RecoveryAckMsg for recovery ack messages, with parsing functions.

Sure, will add `RecoveryAckMsg` wrapper class in this PR for consistency.

python/chilldkg_ref/chilldkg.py
```diff
@@ -1140,7 +1155,7 @@ def participant_recovery_acks_verify(
         RecoveryDataError: If the recovery data is invalid or does not match
             the provided parameters.
         InvalidRecoveryAckError: If any recovery acknowledgment signature is
-            invalid. Note that this does NOT mean the DKG failed
+            malformed or invalid. Note that this does NOT mean the DKG failed
```
real-or-random: I think "malformed" is a suitable term here, but we don't use this anywhere else in similar cases of wrong length. Maybe don't add this now and I suggest that we can reconsider improving the prose here after we solved the "how to deal with invalid message lengths" issue.

real-or-random (commented):
ACK mod this wording nit

python/chilldkg_ref/chilldkg.py
DarkWindman: Okey, agreed!

real-or-random (approved):
ACK 91d5eccb1aa1d4fe376b205f0b63f3048dafc5cf

mllwchrry:
ACK 91d5ecc

PR #125, opened 2026-06-22
More test vectors for coordinator and participant

DarkWindman:
Expands test vector coverage:

- `coordinator_step1`: `t > n`, `pmsgs1` length mismatch, empty `pmsg1`,
  truncated `pmsg1`, infinity hostpubkey.
- `coordinator_finalize`: extra `pmsgs2`, short signature.
- `participant_step1`: infinity hostpubkey.
- `participant_step2`: infinity in `pubnonces` and `sum_coms_to_nonconst_terms`.


python/gen_vector_utils/coordinator.py
```diff
@@ -126,7 +143,60 @@ def generate_coordinator_step1_group(t, n, tc_id_init=0):
...
+        ValueError,
+    )
+    tc_id += 1
+    error_cases.append(
+        {
+            "tcId": tc_id,
+            "pmsg1Indices": list(range(n - 1)),
+            "params": params_asdict(params),
+            "expectedError": error,
+            "comment": "pmsgs1 length does not match hostpubkeys length",
```
mllwchrry:
```suggestion
            "comment": "invalid pmsgs1: fewer entries than participants",
```

python/gen_vector_utils/coordinator.py
```diff
@@ -126,7 +143,60 @@ def generate_coordinator_step1_group(t, n, tc_id_init=0):
...
```
mllwchrry:
```suggestion
            "comment": "invalid pmsgs1: more entries than participants",
```

python/gen_vector_utils/coordinator.py
```diff
@@ -126,7 +143,60 @@ def generate_coordinator_step1_group(t, n, tc_id_init=0):
...
+    long_pmsgs1 = pmsgs1 + [pmsgs1[0]]  # Add an extra entry
+    error = expect_exception(
+        lambda: coordinator_step1(long_pmsgs1, params),
+        ValueError,
+    )
+    tc_id += 1
+    error_cases.append(
+        {
+            "tcId": tc_id,
+            "pmsg1Indices": list(range(n - 1)),
```
mllwchrry:
```suggestion
            "pmsg1Indices": list(range(n)) + [0],
```

It is the same as the short case above. It needs to be fixed to get n + 1 entries.

mllwchrry (commented):
Good job! Found a few issues. 

Also, I noticed a pre-existing issue. `coordinator.py` separates test cases with blank lines, and `participant.py` doesn't. We could fix it in this PR or a separate one.

mllwchrry:
ACK 88f5ab7

python/gen_vector_utils/coordinator.py
```diff
@@ -152,6 +222,50 @@ def generate_coordinator_step1_group(t, n, tc_id_init=0):
         }
     )
 
+    # --- Error test case: pmsg1 (index 0) is empty ---
+    empty_pmsgs1 = b""
+    pmsg1_pool.append(bytes_to_hex(empty_pmsgs1))  # index n + 1
+    invalid_pmsgs1 = [empty_pmsgs1] + pmsgs1[1:]
+    error = expect_exception(
+        lambda: coordinator_step1(invalid_pmsgs1, params),
+        chilldkg.FaultyParticipantError,
```
real-or-random, line 231: Could be done in a separate PR but I wonder if we should have a special except_exception for `Faulty...` errors where we can also specify the faulty index that we expect.

Otherwise, it's easy for us to get it wrong. If our code does it wrongly, then it will end up wronlgy in the test vectors. (Whereas a wrong message is rather cosmetic.)

real-or-random (commented):
ACK mod copilot's comments

It's interesting that I missed the two issues the thing had found. I spend quite some time looking at the code and the indices etc...

python/gen_vector_utils/coordinator.py
DarkWindman: Agreed. That's a really good point that we should include. I can add it in the next PR alongside the new `tcId` logic.

real-or-random (approved):
ACK 9634013d362f7f7cee620835082d8f0005359453

mllwchrry:
ACK 9634013

PR #132, opened 2026-07-14
text: Require authenticity of pubkeys before session

real-or-random:
This is subtle. In principle, checking authenticity could be done at any point in time as long as it happens before sending out signatures in step2: if it's postponed to after step2 and fails for some participant, then someone may have sent money to the threshold pubkey but this participant has aborted and won't participate in signing session. This is a situation we want to avoid.

In other words, if a participant sends a signature, this is also meant as a statement that this participant has completed the protocol *entirely* and is ready to sign according to the rules of the application, which includes that authenticity has been checked.

So in principle, we could allow that checking authenticity happens during the protocol, i.e., after the start of step1 but before the start of step2. But that's not very useful in practice: if the check is manual (e.g., out-of-band eyeballing), then it will anyway take much longer than a protocol step; if it's automatic (e.g., a DB lookup), then there's no reason not to perform it before the protocol.

As a conclusion, it's much simpler to require that checking authenticity happens before the start of the session. This is also defense in depth: if public keys are checked before the start of the session, then we never encrypt secret shares to the attacker. Encrypting to the attacker should not be a problem if we abort later, but not doing it in the first place sounds like a good idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,9 +494,8 @@ and thus the list assigns each participant a unique index in the range of `0` to
 The list of host public keys and the signing threshold `t` together
 form the public *session parameters*, the common input to all participants and the coordinator.
 
-ChillDKG requires that all participants have authentic copies of all other participants' host public keys.[^trust-anchor]
+Each participant **must** ensure to have authentic copies of all other participants' host public keys before the start of the session.[^trust-anchor]
 Authenticity of the host public keys can be verified through pairwise out-of-band comparisons between every pair of participants.
-This verification can occur at any time before the DKG session is finalized, in particular before the start of the session.
 
 [^trust-anchor]: No protocol can prevent man-in-the-middle attacks without this or a comparable assumption.
 Note that this requirement is implicit in other schemes as well.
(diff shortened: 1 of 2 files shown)


DarkWindman:
Agreed! From security perspectives I like the defence in depth approaches. And since it is very simple to implement, just reqiurement before the session, I think we should adopt this!
ACK 6764dd9547917153663602ec568e6aa9225702fb.

mllwchrry:
ACK 6764dd9

README.md
```diff
@@ -494,9 +494,8 @@ and thus the list assigns each participant a unique index in the range of `0` to
 The list of host public keys and the signing threshold `t` together
 form the public *session parameters*, the common input to all participants and the coordinator.
 
-ChillDKG requires that all participants have authentic copies of all other participants' host public keys.[^trust-anchor]
+Each participant must ensure to have authentic copies of all other participants' host public keys before the start of the session.[^trust-anchor]
```
real-or-random:
```suggestion
Each participant **must** ensure to have authentic copies of all other participants' host public keys before the start of the session.[^trust-anchor]
```

real-or-random (commented):
This PR should also update the docstring of `SessionParams`

real-or-random:
Updated

mllwchrry:
Nit (minor inconsistency): README uses "ensure to have", docstring uses "ensure that they have obtained". Could align them for consistency.

real-or-random:
> Nit (minor inconsistency): README uses "ensure to have", docstring uses "ensure that they have obtained". Could align them for consistency.

Fixed. I used the variant in the README.

mllwchrry:
ACK ea08117

DarkWindman:
ACK https://github.com/BlockstreamResearch/bip-frost-dkg/commit/ea08117dab12fe143f9b978edbe8c4dd24bf74ed

PR #136, opened 2026-07-16
Address minor comments by Claude

real-or-random:


python/chilldkg_ref/chilldkg.py
```diff
@@ -8,9 +8,9 @@
 `__all__` list. All other definitions are internal.
 
 In addition to the exceptions documented for each function, all public API
-functions may raise built-in exceptions such as `TypeError` or `ValueError` when
-called with arguments of unexpected structure (e.g., wrong type or wrong
-length). These structural errors are not documented per-function.
+functions may raise exceptions built-in in Python such as `TypeError` or
+`ValueError` when called with arguments of unexpected structure (e.g., wrong
+type or wrong length). These structural errors are not documented per-function.
```
mllwchrry: Let's make it "per function" (without hyphen) for consistency with the README copy.

real-or-random: Thanks, fixed. This was by Claude. I hadn't noticed that it fixed grammar here.

mllwchrry:
ACK f13d1ea

real-or-random:
Sorry, I force-pushed this to rebase. (I wanted to rebase #137 instead...) No changes here except the rebase on master, which was fast-forward.

python/chilldkg_ref/chilldkg.py
```diff
@@ -8,9 +8,9 @@
 `__all__` list. All other definitions are internal.
 
 In addition to the exceptions documented for each function, all public API
-functions may raise built-in exceptions such as `TypeError` or `ValueError` when
-called with arguments of unexpected structure (e.g., wrong type or wrong
-length). These structural errors are not documented per-function.
+functions may raise exceptions built-in in Python such as `TypeError` or
```
DarkWindman:
```suggestion
functions may raise built-in Python exceptions such as `TypeError` or
```

DarkWindman (commented):
ACK, but it would sound better if we rephrase that.

python/chilldkg_ref/chilldkg.py
real-or-random: I agree it's akward and I had this first, but I think it loses precision. "built-in Python exception" doesn't say where it's built-in. The reader could also think it's built-in in ChillDKG. And "built-in" is the [official term](https://docs.python.org/3/library/exceptions.html).

But yeah, it's akward. GPT 5.4 suggested "Python built-in exceptions" (instead of "built-in Python exceptions"). I guess that makes sufficiently clear that these are defined by Python. What do you think?

DarkWindman: Agreed, "Python built-in exceptions" is better!

real-or-random: okay, changed!

DarkWindman:
ACK [fcf7188](https://github.com/BlockstreamResearch/bip-frost-dkg/commit/fcf71885b284b1243c6ef1db18588949f13bc898)

PR #142, opened 2026-07-16
Fix unvalidated pubnonce

### LINKED ISSUE #131: Unvalidated pubnonce breaks the blame guarantee
Found by Claude Fable 5:

```
  The 33-byte pubnonce fields are never validated as curve points: not in
  encpedpop.ParticipantMsg.from_bytes (python/chilldkg_ref/encpedpop.py:171),
  not in encpedpop.CoordinatorMsg.from_bytes (encpedpop.py:219), and not by the
  coordinator. ecdh_libsecp256k1 parses the point and raises a bare ValueError.

  I confirmed with a repro: a faulty participant sends an otherwise well-formed
  pmsg1 whose pubnonce has an off-curve x-coordinate. The honest coordinator
  accepts and relays it, and every honest participant's participant_step2 then
  dies with an empty ValueError from deep inside decaps_multi — not a
  ProtocolError. Consequences:

  - It violates the documented exception contract of participant_step2 (the
  module docstring reserves built-in exceptions for caller-supplied arguments of
  unexpected structure; here the trigger is a remote protocol message).
  - It violates the blame guarantee in the "Blaming Faulty Parties" section: "If
  an honest participant aborts the session, then this participant will blame at
  least one participant or the coordinator." Here the participant aborts with
  no attribution, even though the coordinator holds pmsgs1 that would identify
  the culprit.

  Suggested fix: validate pubnonces as points at parse time — the coordinator
  raising FaultyParticipantError(i) in coordinator_step1, and the participant
  raising FaultyParticipantOrCoordinatorError(i) when parsing cmsg1. recover()
  has the same latent issue (a bare ValueError instead of RecoveryDataError),
  though it's unreachable with a valid certificate, since no honest participant
  can have signed a transcript containing an invalid pubnonce (they'd have
  crashed in step 2 first). Still worth normalizing for defense in depth.
```

The issue seems real. 

But I don't with the suggested fix. I'd rather catch the exception raised by `ecdh_libsecp256k1`. This is cheaper.

DarkWindman:
This PR solves #131 by adding a pubnonce validation at the encpedpop level. Now we parse pubnonces as GE, which means we can validate whether they are on the curve or not.
In the first attack scenario (malicious participant), if participant sends an invalid point, the coordinator can now catch it and raise `FaultyParticipantError(i)`. In the second attack (malicious coordinator), If participant receives pubnonce at step 2 that is not a valid curve point, it raises a `FaultyCoordinatorError`. Additionally, I have added pubnonce validation to the recovery data deserialization along with new test vectors.

One open question: is it correct to just raise `FaultyCoordinatorError` in the second scenario? On one hand, sending invalid values implies the coordinator is malicious. However, it wouldn't be a problem for us to find exactly which pubnonce from the list is invalid and raise `FaultyParticipantOrCoordinatorError(i)` with the index instead. Should we keep the existing implementation or modify it to use the second approach?


real-or-random:
> One open question: is it correct to just raise `FaultyCoordinatorError` in the second scenario? On one hand, sending invalid values implies the coordinator is malicious. However, it wouldn't be a problem for us to find exactly which pubnonce from the list is invalid and raise `FaultyParticipantOrCoordinatorError(i)` with the index instead. Should we keep the existing implementation or modify it to use the second approach?

I had to look what we agreed upon. In https://github.com/BlockstreamResearch/bip-frost-dkg/issues/29 we concluded that in these cases, the coordinator should relay and the participants should check and blame.

I'm not sure if we follow that rule strictly everywhere, but with that rule in mind, I think my suggestion to switch to GE was misguided (sorry! :/ ). That means we should keep bytes. The good thing is that this change will be smaller. 

(I only spent 5 min of thinking on this now, so please check if my reasoning makes sense.)

DarkWindman:
> I had to look what we agreed upon. In https://github.com/BlockstreamResearch/bip-frost-dkg/issues/29 we concluded that in these cases, the coordinator should relay and the participants should check and blame.

Ah, right, I completely forgot about that. 

> That means we should keep bytes. The good thing is that this change will be smaller.

Yes, that makes sense!

python/chilldkg_ref/encpedpop.py
```diff
@@ -107,13 +108,16 @@ def decaps_multi(
...
+            try:
+                pad = ecdh(
+                    seckey=deckey,
+                    my_pubkey=enckey,
+                    their_pubkey=pubnonce,
+                    context=context_,
+                    sending=False,
+                )
+            except ValueError as e:
+                raise FaultyParticipantOrCoordinatorError(sender_idx) from e
```
real-or-random: We usually include an error message.

python/chilldkg_ref/encpedpop.py
```diff
@@ -107,13 +108,16 @@ def decaps_multi(
...
-            )
+            try:
+                pad = ecdh(
+                    seckey=deckey,
+                    my_pubkey=enckey,
+                    their_pubkey=pubnonce,
+                    context=context_,
+                    sending=False,
+                )
+            except ValueError as e:
```
real-or-random, line 119: I think this deserves a comment
```suggestion
            except ValueError as e:
            # Since deckey and enckey are well-formed, error must have been
            # by an invalid pubnonce.
```

python/chilldkg_ref/encpedpop.py
real-or-random: Maybe that's a good opportunity to reorder the imports here.

python/chilldkg_ref/encpedpop.py
```diff
@@ -107,13 +108,18 @@ def decaps_multi(
...
+            # Since deckey and enckey are well-formed, error must have been
```
real-or-random:
```suggestion
            # Since deckey and enckey are well-formed, the error must have been
```

that one is on me

python/gen_vector_utils/participant.py
```diff
@@ -345,10 +345,26 @@ def generate_participant_step2_group(t, n):
...
             "cmsg1": bytes_to_hex(invalid_cmsg1),
             "expectedError": error,
-            "comment": "invalid cmsg1: pubnonces list has an invalid value at index 0",
+            "comment": "invalid cmsg1: pubnonces list has an invalid value at index 0 (own pubnonce)",
         }
     )
-    # --- Error test case: pubnonces list in cmsg1 has an arbitrary value at index 0 ---
+    # --- Error test case: pubnonces list in cmsg1 has an invalid value at index 1 ---
+    invalid_cmsg1_parsed = copy.deepcopy(cmsg1_parsed)
+    invalid_cmsg1_parsed.enc_cmsg.pubnonces[1] = b"\xeb" * 33  # invalid pubnonce
```
real-or-random: Strictly speaking not this PR because we had the value already in another vector: I think a better test is to use an invalid value which has the proper 0x02 or 0x03 prefix but the x coordinate is not on the curve. Or both even.

(The same is true whenever we have a vector with invalid group elements.)

DarkWindman: > Or both even.

You mean invalid prefix and valid (but `x` is not on the curve) or both valid prefixes?

P.S. I think it's the first one after all

real-or-random: I meant "invalid prefix and valid (but `x` is not on the curve)".

> P.S. I think it's the first one after all

Hm, kind of. I looked this up again now. For the hostpubkeys we also have these two:
```
invalid_hostpubkey = b"\x03" + 31 * b"\x00" + b"\x05"  # Invalid x-coordinate
[...]
infinity_hostpubkey = b"\x00" * 33  # Infinite point
```

So there are three failure cases we could cover for a group element:
 - valid prefix but not on curve
 - infinity
 - invalid prefix

I guess it can't be wrong to cover all three of them for both hostpubkeys and pubnonces, and we should just do this.

(nit: Note that `invalid_hostpubkey` has the same x=5 coordinate, but constructs it in a different way. Would be nice to have that consistent. Either way is fine.)

real-or-random:
Looks good to me except for the open conversation on validating group elements.

DarkWindman:
The current version covers all three cases you mentioned (including variants for whether it's our own pubnonce or another participant's, since they raise different errors). I also added one more test for hostpubkey (invalid prefix), similar to the pubnonce tests.

python/chilldkg_ref/encpedpop.py
```diff
@@ -107,13 +108,20 @@ def decaps_multi(
...
+                    their_pubkey=pubnonce,
+                    context=context_,
+                    sending=False,
+                )
+            except ValueError as e:
+                # Since deckey and enckey are well-formed, the error must have been
+                # triggered by an invalid pubnonce.
+                raise FaultyParticipantOrCoordinatorError(
+                    sender_idx, "invalid public nonce"
+                ) from e
```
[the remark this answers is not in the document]
real-or-random: Hm, I believe the comment this clear already. Let's ignore this (unless you feel otherwise @DarkWindman ).

real-or-random (approved):
ACK mod the "dash" nit

python/chilldkg_ref/encpedpop.py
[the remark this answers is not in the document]
DarkWindman: Agreed!

python/gen_vector_utils/participant.py
```diff
@@ -151,11 +151,28 @@ def generate_participant_step1_group(t, n):
             "params": params_asdict(invalid_params),
             "random": bytes_to_hex(random),
             "expectedError": error,
-            "comment": "hostpubkeys list contains an invalid value",
+            "comment": "hostpubkeys list contains an invalid value with invalid prefix",
+        }
+    )
+    # --- Error test case: hostpubkeys list contains a value with an off-curve x-coordinate -—-
```
real-or-random:
```suggestion
    # --- Error test case: hostpubkeys list contains a value with an off-curve x-coordinate ---
```
it's still wrong, I think

DarkWindman: Fixed in 64f4b95c3d599c11efcf562d7181305fc017b6e5. Sorry about that!

real-or-random (approved):
utACK 64f4b95c3d599c11efcf562d7181305fc017b6e5

mllwchrry:
ACK 64f4b95

PR #144, opened 2026-07-21
vss: Allow infinity consistently, add len_bytes()

real-or-random:


python/chilldkg_ref/vss.py
```diff
@@ -37,9 +37,24 @@ class VSSCommitment:
...
 
+    @staticmethod
+    def len_bytes(*, t: int) -> int:
+        return 33 * t
+
+    @staticmethod
+    def from_bytes(b: bytes, *, t: int) -> VSSCommitment:
+        if len(b) != VSSCommitment.len_bytes(t=t):
+            raise ValueError
+        ges = [GE.from_bytes_compressed_with_infinity(b[i : i + 33]) for i in range(0, 33 * t, 33)]
```
mllwchrry: nit: exceeds ruff's line length limit

mllwchrry (commented):
All good, found only one nit.

python/chilldkg_ref/vss.py
real-or-random: Uh, fixed (only on the top commit for simplicity.

`ruff format --check` is clean now. But `ruff check` (ruff 0.16.0) throws 172 errors at me locally. See also on CI: https://github.com/BlockstreamResearch/bip-frost-dkg/actions/runs/30098571795/job/89498631161#step:7:12

This has a green check mark because we suppress the return code... We should fix this.

DarkWindman:
> This has a green check mark because we suppress the return code... We should fix this.

We can fix that in a separate PR and not block this one.

ACK [e21a311](https://github.com/BlockstreamResearch/bip-frost-dkg/commit/e21a311b649d35aeeec5393c1a007923a3bd63c8)

PR #145, opened 2026-07-21
python: Fix blame in participant_finalize

### LINKED ISSUE #143: blaming in participant_finalize
`participant_finalize` raises `FaultyParticipantOrCoordinatorError` when it sees an invalid signature. That's a bit questionable because the coordinator wants to get the DKG result too, so it also check all signatures and aborts if one doesn't match. 

So I think `participant_finalize` should raise `FaultyCoordinatorError` instead.

mllwchrry:
Resolves #143.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1053,9 +1053,8 @@ recovery data to this participant.
 
 *Raises*:
 
-- `FaultyParticipantOrCoordinatorError` - If another known participant or the
-  coordinator is faulty. Make sure to read the above warning, and see
-  the documentation of the exception for further details.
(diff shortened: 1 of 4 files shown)


real-or-random (approved):
utACK d65aa5ab2ba061e6b3c0167de6d666fcd66a6da5

DarkWindman:
ACK [d65aa5a](https://github.com/BlockstreamResearch/bip-frost-dkg/pull/145/commits/d65aa5ab2ba061e6b3c0167de6d666fcd66a6da5)

Issue #138, opened 2026-07-16
More test vectors


real-or-random:
- [ ] Some functions could have more valid vectors (e.g., `participant_step1` has only one valid vector)
 - [ ] We should make sure to cover branches / lines (as far as possible)

DarkWindman:
>  Some functions could have more valid vectors (e.g., participant_step1 has only one valid vector)

Did you have any specific test cases in mind? I was thinking we could add that the result is correct for idx=n-1, or test with different values (e.g., using different randomness).

real-or-random:
> the result is correct for idx=n-1,

I can't follow what you suggest here, but when in doubt, just add it. :D  (I guess we can't have too many vectors.)


> test with different values (e.g., using different randomness).

That's what I had in mind for the "valid" cases.

DarkWindman:
> I can't follow what you suggest here, but when in doubt, just add it. :D (I guess we can't have too many vectors.)

I mean, right now we only test for `idx = 0`, but we could also add `idx = n - 1` to cover the last participant in the list. The logic is that if both edge cases work, it should be correct for all the middle ones too. But maybe that's a stretch.

real-or-random:
> > I can't follow what you suggest here, but when in doubt, just add it. :D (I guess we can't have too many vectors.)
> 
> I mean, right now we only test for `idx = 0`, but we could also add `idx = n - 1` to cover the last participant in the list. The logic is that if both edge cases work, it should be correct for all the middle ones too. But maybe that's a stretch.

I see now. That makes sense. We could even test for all idx. It's just a handful of additional vectors per tested function.

real-or-random:
Infinity points in the VSSCommitments (see #144) would be another good example for "valid" vectors that we could add.

PR #148, opened 2026-07-23
Add new valid test vectors for participant

DarkWindman:
This PR partially addresses #138 by expanding valid test-vector coverage for participant functions.

Changes:

- Per-idx valid cases for `participant_step1`, `participant_step2`, and `participant_finalize`. Previously each group had a single valid case for `idx=0`; now there is one valid case per participant index in the group. 

- Additional aux_rand pattern for `participant_step2`: a valid case with `auxRand = 0x00...00`

- Test harness extension (`tests.py`): valid test cases now support per-case overrides for `hostseckey`, `random`, and `auxRand`


DarkWindman:
It's just a draft for now. I think we can expand on it further

PR #149, opened 2026-07-27
Make ruff happy and really check it on CI

real-or-random:
best reviewed commit-by-commit


mllwchrry:
ACK 9d9509f.

real-or-random:
Let me merge this with a single ACK because it's low-risk and it touches so many lines and thus is likely to interfer with other PRs (i.e., likely to need rebase). We can always adjust if we find issues.

PR #150, opened 2026-07-29
Emphasize usage requirements and improve docstrings

### LINKED ISSUE #56: Highlight usage recommendations

real-or-random:
For the usage requirements, I tried to highlight only these for which the user has sole
responsibility. For example, that an input "must" be 32 bytes is
something that the protocol will check, and so we don't need to
highlight it.

Resolves #56.

### DIFF
--- a/README.md
+++ b/README.md
@@ -917,14 +917,15 @@ Perform a participant's first step of a ChillDKG session.
 
 *Raises*:
 
-- `HostSeckeyError` - If the host secret key is invalid, or if the key does not
-  match any entry of `hostpubkeys`.
+- `HostSeckeyError` - If the host secret key is invalid, or if the key does
+  not match any entry of `hostpubkeys`.
 - `InvalidHostPubkeyError` - If `hostpubkeys` contains an invalid public key.
 - `DuplicateHostPubkeyError` - If `hostpubkeys` contains duplicates.
 - `ThresholdOrCountError` - If `1 <= t <= len(hostpubkeys) <= 2**32 - 1` does
   not hold.
-- `RandomnessError` - If `random` is all zeroes (i.e., `b"\x00" * 32`). This check
-  guards against the case of a malfunctioning random number generator.
+- `RandomnessError` - If `random` is all zeroes (i.e., `b"\x00" * 32`). This
+  check guards against the case of a malfunctioning random number
+  generator.
 
 #### RandomnessError Exception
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -907,30 +909,31 @@ Perform a participant's first step of a ChillDKG session.
 
 - `ParticipantState1` - The participant's session state after this step, to
   be passed as an argument to `participant_step2`. The state **must
-  not** be reused (i.e., it must be passed only to one
+  not* be reused (i.e., it *must** be passed only to one
```
DarkWindman:
```suggestion
  not** be reused (i.e., it **must** be passed only to one
  `participant_step2` call).
```

python/chilldkg_ref/chilldkg.py
DarkWindman, line 659:
```suggestion
not** be reused (i.e., it **must** be passed only to one
```
For consistency with `participant_step1`.

README.md
DarkWindman: Ah, #151 already covers this! Sorry...

real-or-random: > Ah, #151 already covers this! Sorry...

Hm, pretty sure it doesn't.

 I'll fix the first missing `*` here. (I wanted to omit bold in the parantheses because this is just explanatory and doesn't constitute a separate usage requirement.)

real-or-random: Okay, wow. I understand what's going on. What you spotted in #151 is not that the commits in #151 differed, but that the README.md differend from the python docstring (also here in this PR). And this is a bug in the pydoc script, namely here:

https://github.com/BlockstreamResearch/bip-frost-dkg/blob/8dea5c96c1b70d9dc04ef920f4e2f2956b20057a/update-pydoc.sh#L16-L18
This (wrongly) works line by line and thus doesn't work in the specific situation when it encounters the line
```
not** be reused (i.e., it **must** be passed only to one
```
It thinks that "be reused (i.e., it" is supposed to be bold...

Because I wanted to get rid of the highlithing in the "(i..e, ...)" part anyway, I just did that for now.

python/chilldkg_ref/chilldkg.py
real-or-random: Not doing this as per my explanation above.

real-or-random:
Should be ready for review again.

real-or-random:
Force-pushed again. Should be good now, really. :)

python/chilldkg_ref/chilldkg.py
DarkWindman: Agreed! I also think pointing it out once is enough.

DarkWindman:
ACK [dd9cc5d](https://github.com/BlockstreamResearch/bip-frost-dkg/pull/150/commits/dd9cc5de977610586d78c9f72375535961ca4a98)

PR #152, opened 2026-07-29
Align terminology and naming with BIP 445

### LINKED ISSUE #147: Check compatibility and consistency with BIP445
- [ ] The signing BIP says "id" and the DKG BIP says "index": We can rename to the more general "id", in particular because "index" already has a different meaning in BIP446)
 - [x]  The signing BIP says n >= 2 (only in the text) whereas we specified it also for n >= 1: @siv2r suggested to change the text in BIP445.

I assume an LLM can figure all inconsistencies quickly (but I haven't tried yet).

mllwchrry:
Resolves #147.

Improves naming consistency with BIP 445:
- Rename participant "index" to "id", matching BIP 445; "index" now refers only to positions in lists.
- Rename `threshold_pubkey` to `thresh_pk` (and the test vector key `thresholdPubkey` to `threshPk`) to match BIP 445.
- Specify byte lengths and encodings in the `DKGOutput` docstring. These fields are the inputs of BIP 445 signing, which documents their encodings explicitly. Mirroring that completes the alignment.


README.md
```diff
@@ -493,7 +493,7 @@ If there is no dedicated coordinator, one of the participants can act as the coo
 Each participant holds a long-term *host key pair* consisting of a *host public key* and a *host secret key*.
 The `n` participants in a DKG session are represented by an ordered list of their host public keys.
 The list must not contain duplicate entries,
-and thus the list assigns each participant a unique index in the range of `0` to `n - 1` according to the position of their host public key in the list.
+and thus the list assigns each participant a unique id in the range of `0` to `n - 1` according to the position of their host public key in the list.
```
real-or-random:
```suggestion
and thus the list assigns each participant a unique identifier in the range of `0` to `n - 1` according to the position of their host public key in the list.
```

BIP445 uses the term "identifier" in the text and docstrings, and `id` just in code.

real-or-random (commented):
Thanks for taking care of this!

We have a few occurences of `idx` (or variants of it in the code). I think these should also be replaced by `id`. 

It might be worth going through the output of `git grep -i index` and `git grep -i idx` and check if all remaining instances make sense. (I guess there are valid used of "index", e.g., when referring to a Python list.)

real-or-random:
cc @siv2r

real-or-random:
Hmmm, sorry, now that I openend this my proposal to use `id` in the code is not optimal either. The reason is that `id` is a Python built-in function. So our use of `id` as a variable does not make the wrong code (the variable just shadows the built-in function), but shadowing can be confusing to humans and also syntax highligthers etc. (I noticed this when opening a source file in Emacs and saw that `id` is printed in a different color. Ruff even has a [lint for it](https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/ but I assume it's disabled by default.)  

There are a few things we could do about:
 - The "official" recommendation (PEP 8) is to use `id_` in this case. Not too bad, but anyone not deeply into Python will stumble about it when reading the code.
 -  Use `participant_id` or just `participant` instead of `id`. (The exceptions currently use `participant`) 
 -  Ignore the problem

Any opinions here? I'm unsure what the best thing is. I'd say let's stick to `id` for now to get this merged, and we can always change later if we want to.

And this needs rebase, sorry. :/

mllwchrry:
I also noticed the shadowing when doing the rename but decided it was acceptable because nothing calls the built-in `id()`. Since you're revisiting it, I looked at how BIP 445 deals with the same problem: its code never uses bare `id`, always `my_id`/`ids`/`curr_id`. Moreover, it uses `my_id` even where the caller isn't the participant in question (e.g. `PartialSigVerifyInternal`), so the convention is effectively "the id of the participant this computation concerns". We could also adopt this approach.

mllwchrry:
Rebased on master. The `id` naming is easy to change on short notice, so I'm fine either merging as is or applying `my_id`/`participant_id` first if we reach a conclusion.

real-or-random:
> Rebased on master. The `id` naming is easy to change on short notice, so I'm fine either merging as is or applying `my_id`/`participant_id` first if we reach a conclusion.

I don't like `curr_id` (current doesn't say much) or `my_id` (as you say, the "my" doesn't fit everywhere). I slighly lean towards `participant_id`. It's very long for a variable of this kind but [explicit is better than implicit](https://peps.python.org/pep-0020/). And we have also have the `params_id`, so this would distinguish it clearly.

But this make me notice a few tiny issues with the `params_id`:
  - We call it "parameters ID" in the docstring. We should probably be called "parameters identifier" then.
  - It's never mentioned in the text in the usage section but I think it deserves a mention.
  - Maybe `params_hash` would be a better name, after all. This would also remove any kind of confusion with the participant ids... 

Anyway, I don't know. :D Let's just merge this for now.

README.md
siv2r: yes

siv2r:
I used `my_id` and `curr_id` in 445 as a quick dirty fix to dodge the syntax highlighting issue and `id()` function.

`participant_id` reads cleaner to me too. For 445 I'd pair it with `id_i` / `id_j` in the interpolation math:

```python
sign(secnonce, secshare, participant_id, session_ctx)
deterministic_sign(secshare, participant_id, aggothernonce, ...)
partial_sig_verify_internal(psig, participant_id, pubnonce, pubshare, session_ctx)
derive_interpolating_value(ids, id_j)   # for id_i in ids: if id_i == id_j: continue
derive_thresh_pubkey(ids, pubshares)    # for id_i, X_i in zip(ids, pubshares)
```

So `participant_id` wherever a participant is the subject, and `id_i` / `id_j` in the Lagrange math. The list uses `ids`. `signer_id` is an option too, but it won't fit well with the verify functions.

Plain `identifier` is another option, that's what RFC 9591 uses, though for ChillDKG `participant_id` disambiguates better next to `params_id` like you said.

PR #154, opened 2026-07-30
Rename params_id to params_hash and mention in text

real-or-random:
This mitigates the `id` naming thing a bit. `id` still shadows the Python built-in function, but there can't be any confusion with the participant id.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1489,6 +1489,8 @@ The `PATCH` version is incremented for other noteworthy changes (bug fixes, test
   * Remove message classes from the public API; protocol messages are passed as `bytes`.
   * Remove `ParticipantMsgParseError` and `CoordinatorMsgParseError` from the public API.
   * Add `RandomnessError`, `InvalidRecoveryAckError`, and `FaultyParticipantError` to the public API.
+  * Align terminology and naming with the draft of [BIP 445](bip-0445.md), including renaming participant "index" to "identifier".
+  * Rename `params_id` to `params_hash` to avoid confusion with the participant identifier and update the hash tag accordingly.
   * Rename `secp256k1proto` to `secp256k1lab` and separate it as an independent subtree for reuse across projects; see the [upstream repository](https://github.com/secp256k1lab/secp256k1lab).
   * Update the preamble per BIP 3.
 * *0.2.0* (2024-12-19): In addition to various readability improvements to specification and reference implementation, the following major changes were implemented:
--- a/README.md
+++ b/README.md
@@ -787,29 +787,29 @@ A `SessionParams` tuple holds the common parameters of a DKG session.
   keys with the [KeySort algorithm specified in
   BIP 327](bip-0327.mediawiki#key-sorting) to abstract away from the order.
 
-#### params\_id
+#### params\_hash
 
 ```python
-def params_id(params: SessionParams) -> bytes
+def params_hash(params: SessionParams) -> bytes
 ```
 
-Return the parameters ID, a unique representation of the `SessionParams`.
+Return a hash of the session parameters for out-of-band comparison.
 
 In the common scenario that the participants obtain host public keys from
 the other participants over channels that do not provide end-to-end
 authentication of the sending participant (e.g., if the participants simply
 send their unauthenticated host public keys to the coordinator, who is
-supposed to relay them to all participants), the parameters ID serves as a
+supposed to relay them to all participants), the parameters hash serves as a
 convenient way to perform an out-of-band comparison of all host public keys.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1456,6 +1457,8 @@ The `PATCH` version is incremented for other noteworthy changes (bug fixes, test
   * Remove message classes from the public API; protocol messages are passed as `bytes`.
   * Remove `ParticipantMsgParseError` and `CoordinatorMsgParseError` from the public API.
   * Add `RandomnessError`, `InvalidRecoveryAckError`, and `FaultyParticipantError` to the public API.
+  * Align terminology and naming with the draft of [BIP 445](bip-0445.md), including renaming participant "index" to "identifier".
+  * Rename `param_id` to `param_hash` to avoid confusion with the participant identifier.
```
mllwchrry:
```suggestion
  * Rename `params_id` to `params_hash` to avoid confusion with the participant identifier.
```

python/tests.py
```diff
@@ -417,13 +417,13 @@ def test_params_id_vectors():
 
     for test_case in valid_test_cases:
         params = params_from_dict(test_case["params"])
-        expected_id = bytes.fromhex(test_case["expectedParamsId"])
-        assert expected_id == chilldkg.params_id(params)
+        expected_id = bytes.fromhex(test_case["expectedParamsHash"])
```
mllwchrry: Nit: let's maybe call it `expectedParamsHash` or `expectedHash`.

README.md
```diff
@@ -499,7 +499,8 @@ The list of host public keys and the signing threshold `t` together
 form the public *session parameters*, the common input to all participants and the coordinator.
 
 Each participant **must** ensure to have authentic copies of all other participants' host public keys before the start of the session.[^trust-anchor]
-Authenticity of the host public keys can be verified through pairwise out-of-band comparisons between every pair of participants.
+A participant can verify authenticity by comparing every other host public key with the corresponding participant out-of-band,
+or more conveniently, by comparing a short *parameters hash* (a hash of the session parameters) with every other participant out-of-band.
```
mllwchrry, line 503: Would it make sense to add something like "(see the `params_hash` function)" referencing the API here?

mllwchrry (commented):
Two spots still call the hash an identifier: the example's output line (Session parameters identifier: in example.py) and the vector description ("Computes a unique 32-byte identifier …" in gen_vector_utils/session.py). Since the motivation for this PR is removing the id confusion, these should probably say "hash" too.

README.md
real-or-random: fixed

python/tests.py
real-or-random: You mean the variable? Renamed it from `expected_id` to `expected_hash`.

README.md
real-or-random: Maybe, but I didn't do it because we also don't do it for the other functions and data structures (e.g., recovery data).

python/tests.py
mllwchrry: Yes, sorry, I mean variable `expected_hash`.

real-or-random:
rebased

DarkWindman:
ACK [4aa1e77](https://github.com/BlockstreamResearch/bip-frost-dkg/pull/154/commits/4aa1e771537b06011da8de716f9f890c0a2710f7)

mllwchrry:
The PR looks good. But I have two questions to clarify before leaving an ACK:

- What do you think about this one?
> Two spots still call the hash an identifier: the example's output line (Session parameters identifier: in example.py) and the vector description ("Computes a unique 32-byte identifier …" in gen_vector_utils/session.py). Since the motivation for this PR is removing the id confusion, these should probably say "hash" too.
- The changelog entry reads as a pure rename, but the hash tag changed too, so the function output is different. Should we mention this?

real-or-random:
Should be fixed (but I squashed the first two commits to avoid some annoying rebase work).

python/example.py
```diff
@@ -265,7 +265,7 @@ def main():
     print("=== Session parameters ===")
     pphex(params)
     print()
-    print(f"Session parameters identifier: {params_id(params).hex()}")
+    print(f"Session parameters hash : {params_hash(params).hex()}")
```
mllwchrry:
```suggestion
    print(f"Session parameters hash: {params_hash(params).hex()}")
```

mllwchrry (commented):
ACK with one nit.

python/example.py
real-or-random: sorry for that one. fixed.

PR #155, opened 2026-07-30
python: Rename id to participant_id

mllwchrry:
Follow-up to the naming discussion in #152: rename bare `id` to `participant_id` to stop shadowing the Python built-in.


mllwchrry:
Closing in favor of #156.

PR #156, opened 2026-07-30
python: Rename id to participant_id (and similar renames)

real-or-random:
on top of #154, alternative to #155


mllwchrry:
Renaming the exception attribute to `participant_id` is a real improvement I missed. Though I'd keep `faulty_id` and `sender_id` as they were, since "faulty" and "sender" already denote participants. Fine either way though.

python/chilldkg_ref/encpedpop.py
```diff
@@ -272,7 +272,7 @@ class ParticipantState(NamedTuple):
     simpl_state: simplpedpop.ParticipantState
     pubnonce: bytes
     enckeys: list[bytes]
-    id: int
+    particant_id: int
```
DarkWindman:
```suggestion
    participant_id: int
```

python/chilldkg_ref/simplpedpop.py
```diff
@@ -326,7 +328,7 @@ def participant_investigate(
     cinv: bytes,
     partial_secshares: list[Scalar],
 ) -> NoReturn:
-    n, id, secshare, pubshare = error.inv_data
+    n, partipant_id, secshare, pubshare = error.inv_data
```
DarkWindman:
```suggestion
    n, participant_id, secshare, pubshare = error.inv_data
```

python/chilldkg_ref/simplpedpop.py
```diff
@@ -346,7 +348,7 @@ def participant_investigate(
         if not VSSCommitment.verify_secshare(
             partial_secshares[i], partial_pubshares[i]
         ):
-            if i != id:
+            if i != partipant_id:
```
DarkWindman:
```suggestion 
            if i != participant_id:
```

DarkWindman (commented):
Typos

real-or-random:
Will address your comments (and also the CI failure -- I thought I had tested this locally) once #154 is merged.

>  Though I'd keep faulty_id and sender_id as they were, since "faulty" and "sender" already denote participants.

True yeah, I guess I was overdoing it. I'll try to revert but if it ends up in chaos, I'll leave it as it is right now.

real-or-random:
Ready for review. I've also reverted `faulty_id` and `sender_id`.

python/chilldkg_ref/chilldkg.py
mllwchrry: Should these attributes also be renamed to `participant1_id`/`participant_id1` and `participant2_id`/`participant_id2`?

python/vectors/coordinator_finalize_vectors.json
```diff
@@ -112,7 +112,7 @@
                     ],
                     "expectedError": {
                         "type": "FaultyParticipantError",
-                        "participant": 1,
+                        "participant_id": 1,
```
mllwchrry: The renamed attribute makes "participant_id" the first snake_case key in the camelCase vectors.

mllwchrry (commented):
Another thing to mention is that we have stale "indices" wording in the vector description in python/gen_vector_utils/session.py:104. Also, in this description we need to rename `participant` to `participant_id` and likely 'participant1'/'participant2' to the variant with `_id`.

python/vectors/coordinator_finalize_vectors.json
real-or-random: Okay, starting to regret the camelCase thing...  whatever, I've added a fix

python/chilldkg_ref/chilldkg.py
real-or-random: yes, fixed.

DarkWindman:
Should we add a Changelog entry for this, similar to the one for the `params_id` rename?

real-or-random:
I hope I've addressed all comments including the changelog. This entire renaming ended up being more hassle than I hoped for. :/

mllwchrry:
Sorry for finding these one by one instead of in one batch. One more: the descriptions cite 'participant_id' / 'participant_id1'/'participant_id2' as the error-object keys, but with the new camelCase conversion in exception_asdict the actual keys are 'participantId' / 'participantId1'/'participantId2', so the descriptions point at keys that don't exist. Affects the description strings in session.py, coordinator.py, and participant.py.

real-or-random:
> Sorry for finding these one by one instead of in one batch. One more: the descriptions cite 'participant_id' / 'participant_id1'/'participant_id2' as the error-object keys, but with the new camelCase conversion in exception_asdict the actual keys are 'participantId' / 'participantId1'/'participantId2', so the descriptions point at keys that don't exist. Affects the description strings in session.py, coordinator.py, and participant.py.

Oh you couldn't have found this earlier because I introduced it after your previous round of comments.

Fixed, but I've just added a commit to avoid going through the rebase again.

mllwchrry:
ACK 565eaf7.

PR #162, opened 2026-07-30
Clarify Licensing

real-or-random:
BIP recommends having the test vectors under CC0

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) The ChillDKG authors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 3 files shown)


mllwchrry:
ACK 695c29f.

DarkWindman:
ACK https://github.com/BlockstreamResearch/bip-frost-dkg/commit/695c29fda4e9539a8028dd99b343cf4946914e9d.

COPYING
```diff
@@ -0,0 +1,9 @@
+MIT OR CC0-1.0
+
+The auxiliary files of this BIP are provided under one of the following sets of
+terms, at your choice:
+
+ - The MIT License
+   https://opensource.org/license/MIT
+ - CC0 1.0 Universal
+   https://creativecommons.org/publicdomain/zero/1.0/
```
real-or-random: Oh it's wrong, this was supposed to be the file for the vectors dir. Let me fix.

real-or-random:
Force pushed. Now it should match what the copyright section says, no?

COPYING
real-or-random, line 1: note: It's a bit weird here in the dev repo that this file lives in the repo root because then it techncially covers also README.md (but yeah, whatever, I don't care). The idea is that this will end up in the bip-xxxx directory once published and cover only the aux files but not the text.

mllwchrry:
ACK f1bebe7.

